PR #3, opened 2019-02-27
PMM-3281 update docs to reflect consistent mount point

fiowro:


rnovikovP:
pls note  we'll remove mysql and consul from docker too

PR #24, opened 2019-09-13
PMM-4607 http auth for exporters and PMM-4275 architecuture fixes

fiowro:


AlekSi:
Please also resolve conflicts

AlekSi:
One thing that is not right on that picture is arrowheads. They can mean a direction of requests (as opposed to responses) or direction of "main" data (as opposed to control messages). If the intention is to show requests, then there are no requests from Grafana to a user, for example. If the intention is to show the main data flow, then there is mp data flow from Prometheus to exporters.

PR #27, opened 2019-09-18
PMM-4714 installation instructions update

fiowro:


source/install/client-install-apt.rst
```diff
@@ -31,7 +31,7 @@ procedure. |tip.run-all.root|:
    Since PMM 2 is still not GA, you’ll need to use it to enable the experimental component of the original Percona repository::
```
rnovikovP: we are moving to GA already

PR #44, opened 2019-11-19
Gitignore and README.md with installation instructions.

dbazhenov:
I need a text editor.


.gitignore
```diff
@@ -0,0 +1,3 @@
+/build/
+/source/percona-theme/
+.DS*
```
puneet0191: ??

dbazhenov: Because these directories are added to the build. They should not be in the repository.

dbazhenov:
@rnovikovP 
@fiowro  promised me a new text, and I'll wait and send the changes to this PR with a new committee.

PR #52, opened 2019-11-21
PMM-5017 fix pmm-admin add cmdline args omitting

fiowro:


source/manage/client-mongodb-metrics.rst
```diff
@@ -9,14 +9,15 @@ When done, add monitoring as follows:
 
   .. code-block:: bash
 
-     pmm-admin add mongodb  --username=pmm  --password=pmm 127.0.0.1:27017
+     pmm-admin add mongodb  --username=pmm  --password=pmm
```
AlekSi: Double spaces

source/manage/client-mongodb-metrics.rst
```diff
@@ -26,6 +27,7 @@ The output of this command may look as follows:
      Service name: mongo
 
 .. only:: showhidden
+
 	.. code-block:: text
 
 	   $ pmm-admin add mongodb --use-profiler --username=pmm --password=pmm \
```
AlekSi: That command have the same problem. There are more.

fiowro: anything which follows ".. only:: showhidden" (with additional identation) is, well, hidden :) It doesn't go to HTML, it's a portion of old docs which can be useful as a base to enhance the document in future

source/manage/client-mongodb-metrics.rst
fiowro: thanks, fixed!

source/manage/client-mongodb-metrics.rst
AlekSi: So wrong documentation is a good base for enhancement?

fiowro: Yes, as it is a hidden part, which is to be improved to describe PMM2 instead of PMM1 when PMM2 will gain the relevant functionality. Just now it's a sort of placeholder in sources, which shows how to write this part later.

fiowro: But things are different if you notice there something which will never appear in PMM2. In this case we can safely remove the hidden part, as it will never become useful.

fiowro: I remove the part of hidden text which will not be useful in the future (from my point of view)

PR #54, opened 2019-11-22
PMM-4785 fix data retention docs

fiowro:


source/glossary-terminology.rst
```diff
@@ -8,24 +8,20 @@ Terminology Reference
...
-      Depending on available disk space and your requirements, you may
-      need to adjust data retention time.
-
-      You can control data retention by passing the :option:`METRICS_RETENTION`
-      and :option:`QUERIES_RETENTION` environment variables when :ref:`creating
-      and running the PMM Server container <server-container>`.
+      By default, both |prometheus| and :ref:`QAN <QAN>` store time-series data
+      for 30 days.
 
 .. only:: showhidden
```
rnovikovP: Will this be hidden? Why we are hiding options from the uses?

PR #88, opened 2020-01-03
PMM-5215 mention yum update error for ovf image caused by a dependenc…

fiowro:
…y conflict

### DIFF
--- a/source/release-notes/2.2.1.rst
+++ b/source/release-notes/2.2.1.rst
@@ -19,9 +19,10 @@ Improvements and new features
 Fixed bugs
 ==========
 
-* :pmmbug:`5215`: ``yum update`` command was generating errors in the PMM Server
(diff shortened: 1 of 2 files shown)


source/release-notes/2.2.1.rst
```diff
@@ -0,0 +1,27 @@
...
+Improvements and new features
+=============================
+
+Fixed bugs
+==========
+
+* :pmmbug:`5215`: ``yum update`` command was generating errors in the PMM Server
+  2.2.0 OVF image due to a dependency conflict caused by the erroneously
+  pre-installed ``mariadb-libs`` package
+
```
rnovikovP: this makes no sense for the user. Let's rephrase it. 
Note: the  error was visible only in logs; There was no real impact on the user; the main advantage of this fix is to speed up the update by not trying to update  unneeded

PR #94, opened 2020-01-10
PMM-5287 - Fix unclear text, cleaned up instructions and logic as wel…

shoffman-percona:
Some text did not make sense and needed to be cleaned up, several elements were not rendering correctly and some were conflicting with each other so tried to update the page so that the changes would be uniform.


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/pmm-doc?pullRequest=94) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.percona.com:443/percona/pmm-doc?pullRequest=94) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.percona.com:443/check/percona/pmm-doc?pullRequest=94) it.</sub>

source/conf-mysql-requirements.rst
```diff
@@ -110,10 +105,14 @@ follows:
    GRANT SELECT, PROCESS, SUPER, REPLICATION CLIENT, RELOAD ON *.* TO 'pmm'@' localhost' IDENTIFIED BY 'pass' WITH MAX_USER_CONNECTIONS 10;
    GRANT SELECT, UPDATE, DELETE, DROP ON performance_schema.* TO 'pmm'@'localhost';
 
+.. seealso::
+      :ref:`pmm-admin.add-mysql-metrics` - Using the |pmm-admin.add| command
+      to add a monitoring service
+
 
-For more information, run as root
```
shoffman-percona: checked and running the help command as root is not necessary...

source/conf-mysql-requirements.rst
```diff
@@ -110,10 +105,14 @@ follows:
...
+For more information, run:
 |pmm-admin.add|
 |opt.mysql|
-|opt.help|.
+|opt.help|
```
shoffman-percona: There's still something wrong here...on the website this does not render properly but in my local env it does...created https://jira.percona.com/browse/PMM-5288 to track.

shoffman-percona: There's still something wrong here...on the website this does not render properly but in my local env it does...created https://jira.percona.com/browse/PMM-5288 to track.

source/conf-mongodb.rst
```diff
@@ -22,43 +25,45 @@ In |abbr.qan|, you can monitor |mongodb| metrics and |mongodb| queries. Run the
...
 
    Adding a |opt.mongodb-metrics| monitoring service
       :ref:`pmm-admin.add.mongodb-metrics`
+.. The above line does not render...
 
 .. _pmm.qan-mongodb.configuring.profiling.enabling:
 
-`Enabling Profiling <conf-mongodb.html#pmm-qan-mongodb-configuring-profiling-enabling>`_
-========================================================================================
+Enabling Profiling
```
fiowro: Specifying hyperlinks in headers manually is a sort of a trick. It allows us to have titles with human-readable URLs which will not change if we change the title text. It is important to avoid external broken links to our docs. BTW, there is a more natural way to specify links in RST, but for some reason, it makes titles to be rendered with the monospace font and ugly pink background. So these URLs are the least evil.

source/install/docker-setting-up.rst
```diff
@@ -105,12 +106,12 @@ This command does the following:
 * |opt.pmm-server.latest| is the name and version tag of the image
   to derive the container from.
 
-.. _pmm.docker.specific-version:
+.. _specific-version:
```
fiowro: Changing labels may be not the best idea, as some other part may use them to generate the link to this place.  So renaming labels is safe only if label is never met in docs outside of this exact place.

source/conf-mongodb.rst
shoffman-percona: So the root problem is that on the main webpage it's hyperlinked but when you click that link "Enabling Profling" it's not anchored to itself, but instead takes you back to the index which is not desired behavior.  The only one that works as you'd expect/hope is "Setting up the required permissions" above..  I think I treated this heading less as a "linkable bullet" but more as a heading similar to the "Supported versions of MongoDB" earlier on the page. I will update this shortly.

source/install/docker-setting-up.rst
shoffman-percona: Only changed the label because the tag to itself didn't work on it's own page...if I click the text "installing and using specific docker version" it doesn't actually link to that tag correctly...it was the only tag that was prefixed by _pmm.docker.<tagname>...above "Creating and Launching the PMM Server Container" does work so I mimicked that tag.  If you let me know which format is correct I'll recommit with the fix.

rnovikovP:
LGTM, but technical details are better to confirm with @fiowro

PR #97, opened 2020-01-14
PMM-5104 remove cloudwatch grafana datasource from docs

fiowro:


fiowro:
As it turned out, these fixes are false and nothing should be applied.

PR #125, opened 2020-02-17
PMM-5438 alertmanager docs

fiowro:


source/faq.rst
```diff
@@ -216,5 +216,63 @@ You can change the minimum resolution for metrics by the following way:
    when |pmm-server| and |pmm-client| are on different networks,
    or when :ref:`pmm.amazon-rds`.
 
+.. _alertmanager:
+
+How to integrate Alertmanager with PMM?
+================================================================================
+
+The `Alertmanager <https://github.com/prometheus/alertmanager>`_ is the high
```
shoffman-percona: I think it should be "The Alertmanager is a high performance"

source/faq.rst
```diff
@@ -216,5 +216,63 @@ You can change the minimum resolution for metrics by the following way:
...
+performance solution developed by the Prometheus project to handle alerts sent
+by Prometheus. Handling includes deduplicating, grouping, and
+sending alert messages to the specified receiver via e-mail or other
+pre-configured channels. It also takes care of silencing and inhibition of
+alerts if needed.
```
shoffman-percona: Feels like we're leading with what prometheus did and not what we did...I'd suggest flipping the first and second paragraphs so we first talk about what we've done to enable robust alerting from the PMM platform and only then tell people who don't already know what Alertmanager is.

source/faq.rst
```diff
@@ -216,5 +216,63 @@ You can change the minimum resolution for metrics by the following way:
    when |pmm-server| and |pmm-client| are on different networks,
    or when :ref:`pmm.amazon-rds`.
 
+.. _alertmanager:
+
+How to integrate Alertmanager with PMM?
```
shoffman-percona: if this is an FAQ the question should be the most relevant to a wide audience...so this question feels strange...how about "How do I integrate more robust alerting than Grafana native alerts?" or "Is PMM compatible with Prometheus Alertmanager?" or even "Does PMM support Alertmanager?"

This entire blurb is informational but doesn't really answer a question...just describes it...feels more like normal documentation than a frequently asked question.  I'll look at that PR next.

PR #147, opened 2020-04-10
PMM-4074 perf schema privs select only

PaulJacobs-percona:
Removed all but SELECT grants on performance_schema.
Also fixed rst replacements file path.


PaulJacobs-percona:
@fiowro 
The contents of those files are not the same. Each includes distinct files in the parent directory (`../conf-*`). These are included in a few places across the set, I believe. 
The conflict is a result of editing the same file in two separate pull requests.

PR #152, opened 2020-04-15
PMM-5257 Add --socket option and example for proxysql

PaulJacobs-percona:
Described use of `--socket` option and added example of using it (instead of `--port`).

### DIFF
--- a/source/manage/client-proxysql-metrics.rst
+++ b/source/manage/client-proxysql-metrics.rst
@@ -34,14 +34,14 @@ The output of this command may look as follows:
      Service name: ubuntu-proxysql
 
 Beside positional arguments shown above you can specify service name and
-service address with the following flags: ``--service-name``, ``--host`` (the
-hostname or IP address of the service), and either ``--port`` (the port number of the
-service) or ``--socket`` (the UNIX socket path). If both flag and positional argument are present, flag gains higher
-priority. Here is the previous example modified to use these flags for both port and socket connections::
+service address with the following flags: ``--service-name``, and ``--host`` (the
+hostname or IP address of the service) and ``--port`` (the port number of the
+service), or ``--socket`` (the UNIX socket path). If both flag and positional argument are present, flag gains higher
+priority. Here is the previous example modified to use these flags for both host/port or socket connections::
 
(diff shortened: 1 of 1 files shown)


source/manage/client-proxysql-metrics.rst
```diff
@@ -35,12 +35,14 @@ The output of this command may look as follows:
...
-hostname or IP address of the service), and ``--port`` (the port number of the
-service). If both flag and positional argument are present, flag gains higher
-priority. Here is the previous example modified to use these flags::
+hostname or IP address of the service), and either ``--port`` (the port number of the
+service) or ``--socket`` (the UNIX socket path). If both flag and positional argument are present, flag gains higher
+priority. Here is the previous example modified to use these flags for both port and socket connections::
 
      pmm-admin add proxysql --username=pmm --password=pmm --service-name=my-new-proxysql --host=127.0.0.1 --port=3306 
 
+     pmm-admin add proxysql --username=pmm --password=pmm --service-name=my-new-proxysql --host=127.0.0.1 --socket=/var/run/unix_socket
```
askomorokhov, line 44: No, it should be an impossible case: 

We can use host with port:
pmm-admin add proxysql --username=pmm --password=pmm --service-name=my-new-proxysql --host=127.0.0.1 --port=3306

or only socket:

pmm-admin add proxysql --username=pmm --password=pmm --service-name=my-new-proxysql  --socket=/var/run/unix_socket

source/manage/client-proxysql-metrics.rst
```diff
@@ -35,12 +35,14 @@ The output of this command may look as follows:
 
 Beside positional arguments shown above you can specify service name and
 service address with the following flags: ``--service-name``, ``--host`` (the
-hostname or IP address of the service), and ``--port`` (the port number of the
-service). If both flag and positional argument are present, flag gains higher
-priority. Here is the previous example modified to use these flags::
+hostname or IP address of the service), and either ``--port`` (the port number of the
```
askomorokhov, line 38: I guess it should here something like: host with port or socket.

PaulJacobs-percona: I adjusted the sentence and example to show host/port vs socket variants. If you think it needs to be clearer, I will expand the description but it would probably go somewhere else, as this host/port vs socket choices apply to all connectors.

PR #166, opened 2020-05-12
PMM-5243 doc use of --socket option when adding PostgreSQL instance

PaulJacobs-percona:


source/manage/conf-postgres.rst
```diff
@@ -77,6 +77,11 @@ priority. Here is the previous example modified to use these flags::
 
      pmm-admin add postgresql --username=pmm --password=pmm --service-name=postgres --host=127.0.0.1 --port=270175432
```
askomorokhov, line 78: I guess it's better to use real port 5432, as 270175432 is bigger than Unix allow (2^16)

source/manage/conf-postgres.rst
```diff
@@ -77,6 +77,11 @@ priority. Here is the previous example modified to use these flags::
...
 
+.. note:: It is also possible to add a |postgresql| instance using a UNIX socket with
+   just the ``--socket`` flag followed by the path to a socket::
+
+      pmm-admin add postgresql --socket=/var/path/to/postgresql/socket
```
askomorokhov, line 83: And maybe default socket as well `/var/run/postgresql`

PR #168, opened 2020-05-12
PMM-4931 Mention invalid env vars generates a warning message.

PaulJacobs-percona:


source/install/docker-setting-up.rst
```diff
@@ -106,6 +106,8 @@ This command does the following:
 * |opt.pmm-server.latest| is the name and version tag of the image
   to derive the container from.
 
+* A warning message is printed if invalid environment variables are passed in via the command line option ``-e <VARIABLE>``.
```
BupycHuk, line 109: Usually env variables are passed in format `-e <key>=<value>`.
It shows warning message if environment variable key is unknown and fails if environment variable value is invalid.

BupycHuk:
Should we mention that docker container will fail if environment variable value is invalid? @rnovikovP

PR #175, opened 2020-05-18
PMM-5926 Add pmm-agent version to pmm-admin status example output

PaulJacobs-percona:


PaulJacobs-percona:
Needs to be merged when release version for PMM-5926 is known.

PR #182, opened 2020-05-29
PMM-5256 Adding socket info for MongoDB

PaulJacobs-percona:


source/manage/client-mongodb-metrics.rst
```diff
@@ -34,4 +34,9 @@ priority. Here is the previous example modified to use these flags::
 
      pmm-admin add mongodb --username=pmm --password=pmm --service-name=mongo --host=127.0.0.1 --port=27017
 
+.. note:: It is also possible to add a |mongodb| instance using a UNIX socket with
+   just the ``--socket`` flag followed by the path to a socket::
+
+      pmm-admin add mongodb --socket=/var/path/to/mongodb/socket
```
askomorokhov, line 40: If we believe https://github.com/mongodb/mongo/blob/55005a1d4a228174e03862a096029457ca5a7e7c/jstests/noPassthrough/unix_socket.js#L4 default socket should be `/tmp/mongodb-27017.sock`. I guess if we replace dummy socket path with value by the default, the users with usual installation need just copy/paste to get it worked.

PR #186, opened 2020-06-10
PMM-5563 annotations

PaulJacobs-percona:
Complete rework of annotations section. Update of pmm-admin man page. Removal of missing files from index (to reduce Sphinx warnings).

### DIFF
--- a/source/manage/client-annotations.rst
+++ b/source/manage/client-annotations.rst
@@ -20,25 +20,27 @@ They can be set globally or for specific nodes or services.
 
 .. rubric:: USAGE
 
-``pmm-admin annotate <annotation> [--tags <tags>]``
-
-``pmm-admin annotate node <annotation> [--tags <tags>] [--node-name=<node>]``
-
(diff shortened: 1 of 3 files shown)


source/manage/client-annotations.rst
```diff
@@ -1,74 +1,59 @@
...
 
-`Marking Important Events with Annotations <client-annotations.html#application-event-marking>`_
-------------------------------------------------------------------------------------------------------------
+``--node``
+   Annotate the current node or that specified by ``--node-name``.
 
-Some events in your application may impact your database. Annotations
-visualize these events on each dashboard of |pmm-server|.
+``--service``
+   Annotate the service running on the current node. If there is more than one service, annotate that specified by ``--service-name``.
```
JiriCtvrtka, line 34: Annotate all services running on the current node. I am not sure if it is clear without word all. Also if there is more services on current node and also you define --service-name then only service with that name will be annotated.

PaulJacobs-percona: Agreed.

PR #190, opened 2020-06-16
PMM-6073 QAN home page rework & update

PaulJacobs-percona:
Merged separate pages into one to remove duplication and improve readability.


source/index-qan.rst
```diff
@@ -1,14 +1,6 @@
-================================================================================
-Using PMM Query Analytics
-================================================================================
+:orphan:
```
PaulJacobs-percona, line 1: Some of these index files were missing so the index level was removed and all content merged into one file.

source/qan-intro.rst
```diff
@@ -1,40 +1,268 @@
+.. raw:: html
```
PaulJacobs-percona, line 1: This is the consolidated file with merged content.

source/reference/mongodb-metrics.csv
```diff
@@ -0,0 +1,6 @@
+Metric Name,Metric Key,Unit Type,Relation,Relation Units Type
```
PaulJacobs-percona, line 1: Metrics for each DB are stored as a CSV file and imported for ease of maintenance. The info comes from confluence and is incomplete there.

PR #198, opened 2020-06-29
PMM-5324 metrics resolution

PaulJacobs-percona:
Consolidated info for "What resolution is used for metrics?" FAQ item.
General improvements to readability for other FAQ items.
Reformatting reST to PMM standards.

### DIFF
--- a/source/faq.rst
+++ b/source/faq.rst
@@ -56,11 +56,8 @@ A minimum of 100 MB of storage is required for installing the PMM Client package
 How to control data retention for PMM?
 **************************************
 
-By default, both Prometheus and QAN store time-series data for 30 days.
-
-Depending on your available disk space and requirements, you may need to adjust the data retention time.
-
-You can control data retention by the following way.
+By default, PMM stores time-series data for 30 days.
+Depending on your available disk space and requirements, you may need to adjust the data retention time:
 
 #. Select the PMM Settings dashboard in the main menu.
 
--- a/source/faq.rst
+++ b/source/faq.rst
@@ -7,7 +7,7 @@
 Frequently Asked Questions
 ##########################
 
-.. contents:: Contents
+.. contents::
    :local:
    :depth: 1
 
@@ -184,9 +184,7 @@ The default values are:
 How to set up Alerting in PMM?
 ******************************
 
-When a monitored service metric reaches a defined threshold, PMM Server can trigger alerts for it either using the Grafana Alerting feature
-
-or by using an external Alertmanager  (a high-performance solution developed by the Prometheus project to handle alerts sent by Prometheus).
+When a monitored service metric reaches a defined threshold, PMM Server can trigger alerts for it either using the Grafana Alerting feature or by using an external Alertmanager (a high-performance solution developed by the Prometheus project to handle alerts sent by Prometheus).
 
 With both methods you must configure alerting rules that define conditions under which an alert should be triggered, and the channel used to send the alert (e.g. email).
 
--- a/source/faq.rst
+++ b/source/faq.rst
@@ -1,6 +1,3 @@
-.. include:: /.res/replace.txt
-
-
 .. _faq:
 
 ##########################
@@ -180,15 +177,14 @@ The default values are:
 .. _alertmanager:
 .. _how-to-setup-alerting-with-grafana:
 
-******************************
-How to set up Alerting in PMM?
-******************************
+********************************
+How do I set up Alerting in PMM?
+********************************
 
-When a monitored service metric reaches a defined threshold, PMM Server can trigger alerts for it either using the Grafana Alerting feature or by using an external Alertmanager (a high-performance solution developed by the Prometheus project to handle alerts sent by Prometheus).
-
-With both methods you must configure alerting rules that define conditions under which an alert should be triggered, and the channel used to send the alert (e.g. email).
-
+When a monitored service metric reaches a defined threshold, PMM Server can trigger alerts for it either using the Grafana Alerting feature or by using an external Alertmanager, a high-performance solution developed by the Prometheus project to handle alerts sent by Prometheus.
 (Both are advanced features that require knowledge of third-party tools.)
+
+With these methods you must configure alerting rules that define conditions under which an alert should be triggered, and the channel used to send the alert (e.g. email).
 
 Grafana Alerts are already integrated into PMM Server and may be simpler to get set up.
 
@@ -208,11 +204,11 @@ Alerting in Grafana allows attaching rules to your dashboard panels.
 
    - `PMM Alerting with Grafana: Working with Templated Dashboards <https://www.percona.com/blog/2017/02/02/pmm-alerting-with-grafana-working-with-templated-dashboards/>`_
 
-***************************************************************************
-How do I use a custom Prometheus configuration file inside of a PMM Server?
-***************************************************************************
+**********************************************************************
+How do I use a custom Prometheus configuration file inside PMM Server?
+**********************************************************************
 
-Normally, PMM Server fully manages `Prometheus configuration file <https://prometheus.io/docs/prometheus/latest/configuration/configuration/>`_.
+Normally, PMM Server fully manages the `Prometheus configuration file <https://prometheus.io/docs/prometheus/latest/configuration/configuration/>`_.
 
 However, some users may want to change the generated configuration to add additional scrape jobs, configure remote storage, etc.
 
--- a/source/faq.rst
+++ b/source/faq.rst
@@ -46,7 +46,7 @@ The minimum memory requirement is 2 GB for one monitored database node.
 Any modern 64-bit Linux distribution. It is tested on the latest versions of Debian, Ubuntu, CentOS, and Red Hat Enterprise Linux.
 
 A minimum of 100 MB of storage is required for installing the PMM Client package.  With a good connection to PMM Server, additional storage is not required.  However, the client needs to store any collected data that it cannot dispatch immediately, so additional storage may be required if the connection is unstable or the throughput is low.
-
+(Caching only applies to Query Analytics data; Prometheus data is never cached on the client side.)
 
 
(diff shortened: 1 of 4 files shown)


source/faq.rst
```diff
@@ -1,264 +1,219 @@
...
+
+.. rubric:: PMM Client
 
-Minimum memory is 2 GB for one monitored database node, but it is not
-linear when you add more nodes.  For example, data from 20 nodes
-should be easily handled with 16 GB.
+Any modern 64-bit Linux distribution. It is tested on the latest versions of Debian, Ubuntu, CentOS, and Red Hat Enterprise Linux.
 
-.. rubric:: |pmm-client|
+A minimum of 100 MB of storage is required for installing the PMM Client package.  With a good connection to PMM Server, additional storage is not required.  However, the client needs to store any collected data that it cannot dispatch immediately, so additional storage may be required if the connection is unstable or the throughput is low.
```
michaelcoburn, line 48: You may want to be explicit that caching applies to Query Analytics data only, as Prometheus data is never cached on the client side.

source/faq.rst
```diff
@@ -1,264 +1,219 @@
...
 .. _data-retention:
+.. _how-to-control-data-retention-for-pmm:
 
+**************************************
 How to control data retention for PMM?
-================================================================================
+**************************************
 
-By default, both |prometheus| and QAN store time-series data for 30 days.
+By default, both Prometheus and QAN store time-series data for 30 days.
```
michaelcoburn, line 59: I thought we stopped using QAN in favour of Query Analytics?

source/faq.rst
```diff
@@ -1,264 +1,219 @@
...
-What privileges are required to monitor a |mysql| instance?
-================================================================================
+*********************************************************
+What privileges are required to monitor a MySQL instance?
+*********************************************************
+
+.. code-block::
+
+   GRANT SELECT, PROCESS, SUPER, REPLICATION CLIENT, RELOAD ON *.* TO 'pmm'@'localhost';
+   GRANT SELECT ON performance_schema.* TO 'pmm'@'localhost';
```
michaelcoburn, line 94: This line is redundant as we already grant SELECT ... *.* in the earlier privilege. I suggest you remove it.

source/faq.rst
```diff
@@ -1,264 +1,219 @@
...
+*****************************************
 Can I monitor multiple service instances?
-================================================================================
+*****************************************
 
-Yes, you can add multiple instances of |mysql| or some other service to be
-monitored from one |pmm-client|. In this case, you will need to provide a
-distinct port and socket for each instance, and specify a unique name for each
-instance (by default, it uses the name of the |pmm-client| host).
+You can add multiple instances of MySQL or some other service to be monitored from one PMM Client. In this case, you must provide a unique port or socket for each instance, and specify a unique name for each.  (If a name is not provided, PMM uses the name of the PMM Client host.)
```
michaelcoburn, line 104: or IP address for a multi homed instance.  There's three optiosn for running pmm2-client against multiple local instances.

source/faq.rst
```diff
@@ -1,264 +1,219 @@
...
+With these methods you must configure alerting rules that define conditions under which an alert should be triggered, and the channel used to send the alert (e.g. email).
 
-You can make PMM Server trigger alerts when your monitored service reaches some thresholds in two ways:
+Grafana Alerts are already integrated into PMM Server and may be simpler to get set up.
 
-* using `Grafana Alerting feature <https://grafana.com/docs/grafana/latest/alerting/rules/>`_,
-* using external `Alertmanager <https://github.com/prometheus/alertmanager>`_ (a
-  high-performance solution developed by the Prometheus project to handle alerts sent
-  by Prometheus).
+Alertmanager allows the creation of more sophisticated alerting rules and can be easier to manage installations with a large number of hosts. This additional flexibility comes at the expense of simplicity and requires advanced knowledge of Alertmanager rules.
```
michaelcoburn, line 191: Maybe a quibble but why do we keep refering to "advanced knowledge". I think that's initimidating language, and just knowing how to write alerting rules to me doesn't seem advanced. I'd drop it in order to make the conversation more approachable to new users.

source/faq.rst
```diff
@@ -1,264 +1,219 @@
...
 
-Grafana Alerts are already integrated into PMM Server and may be simpler to get set up,
-while Alertmanager allows the creation of more sophisticated alerting rules and
-can be easier to manage installations with a large number of hosts; this
-additional flexibility comes at the expense of simplicity and requires advanced
-knowledge of Alertmanager rules. Currently Percona cannot offer support for
-creating custom rules so you should already have a working Alertmanager instance
-prior to using this feature, however we are working hard to bring an integrated
-Alertmanager solution to make rule generation easy!
+   Percona cannot offer support for creating custom rules so you should already have a working Alertmanager instance prior to using this feature. (However, we are working hard to bring an integrated Alertmanager solution to make rule generation easy.)
```
michaelcoburn, line 197: We actually do offer custom rules if you are a Percona Customer. I would explicitly call this out as it is a feature of our offering.

source/manage/server-admin-gui.rst
```diff
@@ -24,19 +24,28 @@ Press *Apply changes* to store any changes.
 Metrics resolution
 ******************
 
-PMM collects metrics at three different time intervals.
-
+Metrics are collected at three intervals representing low, medium and high resolutions.
```
michaelcoburn, line 27: You don't define what is meant by 'resolutions' - but we use the word a ton. I would encourage yo uto say "scraping intervals" or "frequency of data collection attempts" or something else to ensure the idea is communicated.

source/manage/server-admin-gui.rst
```diff
@@ -24,19 +24,28 @@ Press *Apply changes* to store any changes.
...
+
+- Low: 60 seconds
+- Medium: 10 seconds
+- High: 5 seconds
+
+The *Metrics Resolution* slider lets you choose from three preset combinations of intervals corresponding to high, medium, and low resolution (short, medium, and long collection periods).
+
+The slider tool-tip shows the collection time corresponding to each resolution setting.
+
+- Setting the slider to *Low* increases the time between collection, resulting in low-resolution metrics.
```
michaelcoburn, line 40: "resulting in data being collected less frequently, thus reducing the storage cost but at a tradeoff of resolution" or something like thist

source/manage/server-admin-gui.rst
```diff
@@ -58,13 +67,13 @@ We do not gather anything that would make the system identifiable, but the follo
...
+The first telemetry reporting of a new PMM Server instance is delayed by 24 hours to allow sufficient time to disable the service for those that do not wish to share any information.
 
 There is a landing page for this service, available at `check.percona.com <https://check.percona.com>`_, which clearly explains what this service is, what it’s collecting, and how you can turn it off.
 
-.. note:: The `Grafana internal reporting feature <https://grafana.com/docs/grafana/latest/installation/configuration/#reporting-enabled>`_ is currently **not** managed by PMM. If you want to turn it, you need to go inside the PMM Server container and `change configuration <https://grafana.com/docs/grafana/latest/installation/configuration/#reporting-enabled>`_ after each update.
+The `Grafana internal reporting feature <https://grafana.com/docs/grafana/latest/installation/configuration/#reporting-enabled>`_ is currently **not** managed by PMM. If you want to turn it, you need to go inside the PMM Server container and `change configuration <https://grafana.com/docs/grafana/latest/installation/configuration/#reporting-enabled>`_ after each update.
```
michaelcoburn, line 74: "if you want to turn it" ? what are you trying to say here?

also why do we call it "grafana internal reporting feature", they call it "anonymous usage statistics" so we should call it that as well

source/faq.rst
PaulJacobs-percona: @rnovikovP Is that true?

rnovikovP: Why on earth do we need 2 objects/terms here if this is PMM who stores data?  there is no option to set the different retention periods for them separately. 
As for  QAN vs Query Analytics  - I'll belive Michael in this case )) it's better to avoid TLA  (Three-letter acronym)

source/faq.rst
PaulJacobs-percona: Thanks. I'd copied it from https://www.percona.com/doc/percona-monitoring-and-management/2.x/concepts/services-mysql.html#pmm-conf-mysql-user-account-creating
where I will change it too.

source/faq.rst
PaulJacobs-percona: Not a quibble, you're right. I guess the original intent was as a gentle warning.

source/manage/server-admin-gui.rst
PaulJacobs-percona: Good spot. We've had some chats about this terminology already but no clear ... resolution. I'm concerned that the terms are not intuitive and I don't want the docs to keep having to explain terms because they are different to the UX. I will chat with our new UI guy and see if we can come up with a clearer explanation, or suggest a new UI metaphor.

source/manage/server-admin-gui.rst
PaulJacobs-percona: Can I instead add to the note under this section a mention of the tradeoff: hi-res = more resources but more accuracy, low-res = less resources, less accuracy.

source/manage/server-admin-gui.rst
PaulJacobs-percona: I missed that. I'm guessing it was meant to be 'if you want to turn it on'. 
I'll also change the name of the feature.

source/faq.rst
PaulJacobs-percona: No idea why both Prometheus and Query Analytics are mentioned in this context. Perhaps it was to explain where the data is stored, by which components. I'll rephrase it to just say PMM.

rnovikovP: it was there because previously we had two different variables to ser retention for Prom (30 days by default ) and for QAN (7 days by default )

PR #202, opened 2020-07-15
PMM-5697 Supported Ubuntu

PaulJacobs-percona:
Updated list of supported Linux.
Removed replacements, includes, asciicinema snippets.
Tidied and cleaned rst syntax according to style guidelines.

### DIFF
--- a/source/install/client-install-apt.rst
+++ b/source/install/client-install-apt.rst
@@ -1,56 +1,66 @@
 .. _install-client-apt:
 
-Installing DEB packages using apt-get
-================================================================================
+#########################################
+Installing DEB packages using ``apt-get``
+#########################################
 
-If you are running a DEB-based |linux| distribution, use the |apt| package
-manager to install |pmm-client| from the official Percona software repository.
+If you are running a DEB-based Linux distribution, you can use the ``apt`` package
+manager to install PMM client from the official Percona software repository.
 
-|percona| provides :file:`.deb` packages for 64-bit versions of the following
+Percona provides ``.deb`` packages for 64-bit versions of the following
 distributions:
 
-.. include:: ../.res/contents/list.pmm-client.supported-apt-platform.txt
+* Debian 8 (jessie)
+* Debian 9 (stretch)
+* Debian 10 (buster)
+* Ubuntu 16.04 LTS (Xenial Xerus)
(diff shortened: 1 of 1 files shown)


source/install/client-install-apt.rst
```diff
@@ -1,56 +1,66 @@
...
-manager to install |pmm-client| from the official Percona software repository.
+If you are running a DEB-based Linux distribution, you can use the ``apt`` package
+manager to install PMM client from the official Percona software repository.
 
-|percona| provides :file:`.deb` packages for 64-bit versions of the following
+Percona provides ``.deb`` packages for 64-bit versions of the following
 distributions:
 
-.. include:: ../.res/contents/list.pmm-client.supported-apt-platform.txt
+* Debian 8 (jessie)
```
EvgeniyPatlan, line 13: it is EOL since June 30. https://wiki.debian.org/LTS

PaulJacobs-percona:
> I think we should discuss removing Debian 8 builds and support

Who is 'we' and when will that conversation happen? Does it block this PR?

source/install/client-install-apt.rst
```diff
@@ -1,56 +1,65 @@
...
-If you are running a DEB-based |linux| distribution, use the |apt| package
-manager to install |pmm-client| from the official Percona software repository.
+If you are running a DEB-based Linux distribution, you can use the ``apt`` package
+manager to install PMM client from the official Percona software repository.
 
-|percona| provides :file:`.deb` packages for 64-bit versions of the following
+Percona provides ``.deb`` packages for 64-bit versions of the following
 distributions:
 
-.. include:: ../.res/contents/list.pmm-client.supported-apt-platform.txt
```
EvgeniyPatlan, line 12: I have one more question. As I remember it was decided some time ago that all products should point to supported platforms page and do not mention supported platforms on it's own page. Why you decided to put these supported platforms here?

PaulJacobs-percona: Yes, I have seen that too. This section is a carry-over from PMM1. Also, there is a conflict in that the supported platforms page still lists Debian 8 support for PMM2 as active. I want to change the documentation structure so there is a 'prerequisites' section for all software, not individual pages as they are at the moment. So this is a quick-fix for accuracy.

PR #230, opened 2020-08-03
PMM-5608 Change QAN mongodb permissions role/user example code

PaulJacobs-percona:


PaulJacobs-percona:
@Nailya Two versions of the same change. Please check or nominate someone to do it. Thanks.

PR #245, opened 2020-08-25
Additional use cases for annotations

rnovikovP:


PaulJacobs-percona (approved):

source/manage/client-annotations.rst
```diff
@@ -42,7 +42,50 @@ They can be set globally or for specific nodes or services.
...
+
+Flags may be combined as shown in the following examples.
+
+``--node``
+    current node
+
+``--node-name``
+    node with name
+
+``--node --node-name``
```
rnovikovP, line 57: maybe something like :
--node --node-name=SOME_NODE_NAME  

 ?

source/manage/client-annotations.rst
```diff
@@ -42,7 +42,50 @@ They can be set globally or for specific nodes or services.
...
+    node with name
+
+``--node --service-name``
+    current node and service with name
+
+``--node --node-name --service-name``
```
rnovikovP, line 63: maybe we can wrap this and other cases to say "if  node  or service name is specified  - they will be used  instead of other parameters "

PR #257, opened 2020-09-16
Update qan-intro.rst

shoffman-percona:
Not sure if there are version requirements for this so needs a little bit of research before posting, but wanted to see if this was a more effective way of getting that into the pipeline for resolution than just creating a Jira ticket!

### DIFF
--- a/source/qan-intro.rst
+++ b/source/qan-intro.rst
@@ -20,7 +20,7 @@ The *Query Analytics* dashboard shows how queries are executed and where they sp
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


PaulJacobs-percona:
> Not sure if there are version requirements for this so needs a little bit of research before posting, but wanted to see if this was a more effective way of getting that into the pipeline for resolution than just creating a Jira ticket!

Jira can be bypassed. Github's Issues are an option too.
(BTW Can you replicate this change in md files too please?)

PR #270, opened 2020-10-22
PMM-5974 New DBaaS instructions

PaulJacobs-percona:
Notes:
1. Haven't shown new red delete button because screenshots clash with change of Kubernetes add/register labeling. New images will be added when both FBs are merged.
2. These docs are not visible (not included in navigation).


lunaticusgreen (approved):
Looks good

PR #276, opened 2020-11-02
Update faq.md

shoffman-percona:
Waiting for the git id's for Brandon and Carlos so they can help craft just the right wording.  FAQ seemed to make the most sense but if you think there's a better place, your call @PaulJacobs-percona


PaulJacobs-percona (approved):

PaulJacobs-percona:
> Waiting for the git id's for Brandon and Carlos so they can help craft just the right wording. FAQ seemed to make the most sense but if you think there's a better place, your call @PaulJacobs-percona

There isn't a better place at the moment, but after the ToC reorg there will be a natural home for it. (How to->Upgrade)

PR #299, opened 2020-11-20
PMM-6482 Old + new with JS redirects

PaulJacobs-percona:


PaulJacobs-percona (commented):
LGTM

PaulJacobs-percona:
Reviewers: You can't check each file so here's a summary. 
The new structure is merged with the old. Existing files are 'empty' except for a javascript redirect to the new page. These prevent 404s until the server 301s are in place. Then they can be deleted.
We need this in before 2.12 so that the ever-expanding DBaaS page has a natural home and we don't have to duplicate edits in old/new structures.

rnovikovP (approved):
assume this is based on the document we havefor redirects

PaulJacobs-percona:
> assume this is based on the document we havefor redirects

Yes it is.

shoffman-percona (approved):
I love the structure and the possibility of adding more to it as well!  this is probably out of scope of the change you're asking for but when I click a "parent" section (i.e. setting up) I have an expectation of it both loading any relevant content in the main window but also having the same effect as expanding the tree but when I click setup it just takes me to the setup page and doesn't expand the TOC to see the aggregated contents.

PR #304, opened 2020-11-27
PMM-6959 Release Notes 2.12.0

PaulJacobs-percona:


shoffman-percona (commented):
I just went through Jira and updated a TON of summaries so how hard is it to regen this list with the latest jira summary? (and should I be doing it that way or another?

PaulJacobs-percona:
> I just went through Jira and updated a TON of summaries so how hard is it to regen this list with the latest jira summary? (and should I be doing it that way or another?

Changing in Jira is the right way. jirabot rewrites the file.

rnovikovP (approved):

PR #306, opened 2020-12-03
Admonition styles

PaulJacobs-percona:
Styles for admonitions need to use the bootstrap names instead of MkDocs's.


PaulJacobs-percona:
Pending changes to some JS in Drupal to make Bootstrap alerts non-modal by default

PR #342, opened 2021-01-01
PMM-7249 Integrated Alerting

PaulJacobs-percona:
- Clarifying the distinction between (new) Integrated Alerting, (old) Prometheus Alerting, (Grafana) Alert functionality.
- Flow diagram (currently hidden) to show what to do to get IA set up.


rnovikovP:
LGTM  )

docs/how-to/configure.md
```diff
@@ -178,26 +178,30 @@ A brief message will confirm the creation of your new account and you may now lo
...
+- *Hello*: The default hostname to identify to the SMTP server.
+- *Identity*: SMTP Auth using PLAIN.
+- *Secret*: SMTP Auth using CRAM-MD5.
 
 ### Slack
 
-- URL
+Settings for Slack notifications:
+
+- *URL*: The API URL to use for Slack notifications.
```
palash25: slack webhook URL is the preferred term

docs/using/alerting.md
```diff
@@ -1,44 +1,48 @@
...
 ---
 
-!!! alert alert-warning "Warning"
-    Integrated alerting is a technical preview and is subject to change.
-
-## Alerts
+## Definitions
 
-An alert has a name, a summary, a description, and a definition as a set of rules.
+- Alerts happen when their criteria (*alert rules*) are met; an *alert* is the result of an *alert rule* firing.
```
palash25: > *alert* is the result of an *alert rule* firing

more like `an alert is the result of an alert rule expression evaluating to true`, because for an alert may not always be in a firing state. For e.g. if we use the MySQL down template to create an alert rule and enter 60s as the duration, if at some point in time the MySQL db goes down for let's say just 30s and alert is generated but in a pending state, it will only be in a firing state if it has existed for a duration longer than the one provided in our alert rule that is 60s in this example.

palash25: >  Alerts happen

nitpick: I think "Alerts are generated" would be a better way to phrase this.

docs/using/alerting.md
```diff
@@ -1,44 +1,48 @@
...
+- Alert rules are based on *alert rule templates*. We provide a default set of templates. You can use them as-is, or modify and extend them to create new ones.
```
palash25: > You can use them as-is, or modify and extend them to create new ones.

we can't modify our built-in templates, our built-in templates are converted to generated go code which is the part of the managed binary.

the users can modify the templates that they write themselves but not the ones provided by Percona.

docs/how-to/configure.md
```diff
@@ -178,26 +178,30 @@ A brief message will confirm the creation of your new account and you may now lo
 !!! alert alert-info "Note"
     This tab appears only when *Advanced Settings* --> *Integrated Alerting* is on.
 
-Alertmanager communications settings.
-
+Global communications settings for [Integrated Alerting](../using/alerting.md).
 
 ![](../_images/PMM_Settings_Communication.jpg)
 
+Integrated Alerting uses a separate instance of Alertmanager. The descriptions for the settings here are reproduced from [Prometheus Alertmanager configuration](https://prometheus.io/docs/alerting/latest/configuration/).
```
palash25: > Integrated Alerting uses a separate instance of Alertmanager

let's be a bit verbose here. The separate alertmanager used for integrated alerting is already run by pmm-managed so we should add the information here so that the user doesn't get confused and think about starting their own separate instance of alertmanager to use Integrated Alerting.

"Integrated Alerting uses a separate instance of Alertmanager run by pmm-managed or the pmm server (whichever term is used in the docs)"

palash25 (changes requested):

palash25:
i don't see these changes when I try to look at the unified diff in the "Files changed" tab, I had to go to the specific commit 0b6501fe  to review it. Did the branch history change @PaulJacobs-percona ?

PaulJacobs-percona:
> i don't see these changes when I try to look at the unified diff in the "Files changed" tab, I had to go to the specific commit 0b6501fe to review it. Did the branch history change @PaulJacobs-percona ?

Perhaps it was originally for 2.13 branch.

PR #351, opened 2021-01-12
PMM-6309 replace or remove offensive terms

PaulJacobs-percona:
Many cannot be removed as they are technical elements (variables, commands, etc). Others live in links to docs (will be resolved with PMM-7046)


docs/js/version-select.js
```diff
@@ -0,0 +1,58 @@
...
+  // template refers to...
+  var ex = new RegExp("/?css/version-select.css$");
+  var sheet = document.querySelector('link[href$="version-select.css"]');
+
+  var ABS_BASE_URL = sheet.href.replace(ex, "");
+  var CURRENT_VERSION = ABS_BASE_URL.split("/").pop();
+
+  function makeSelect(options, selected) {
+    var select = document.createElement("select");
+//    select.classList.add("form-control");
```
atymchuk, line 12: Redundant code

mkdocs.yml
```diff
@@ -1,5 +1,5 @@
 site_name: Percona Monitoring and Management
-site_url: https://www.percona.com/doc/percona-monitoring-and-management/2.x/
+#site_url: https://new.percona.com/doc/percona-monitoring-and-management/2.x/
```
atymchuk, line 2: Would it work if `site_url` is commented out?

atymchuk (changes requested):

mkdocs.yml
PaulJacobs-percona: Yes it does and means it works on both new.percona.com and www.percona.com

docs/js/version-select.js
PaulJacobs-percona: Residue from testing

PR #354, opened 2021-01-14
configuration hint for resource consumption

shoffman-percona:
since we don't do any pre-validation, makes sense to tell the user up front how to know if they're k8s cluster has enough resources to start their desired DB cluster config.


PaulJacobs-percona (approved):
Moved to top of section. Made into 'Tip' (should be green). Reworded slightly...

browseman (approved):

PR #355, opened 2021-01-15
Add a note about Activation the feature

rnovikovP:


PaulJacobs-percona:
This is already mentioned in 'Prerequisites' a few lines down.

PR #361, opened 2021-01-22
PMM-5765 External serverless

PaulJacobs-percona:


BupycHuk (approved):

BupycHuk (changes requested):
I don't see url or address fields, one of them are required

PaulJacobs-percona:
> I don't see url or address fields, one of them are required

The usage under SYNOPSIS shows an [ADDRESS] is required. This section is just for [FLAGS].

PR #363, opened 2021-01-24
Migrate 3P github actions

atymchuk:


.github/workflows/build.yml
```diff
@@ -9,36 +9,43 @@ on:
...
+          git config --global url."https://percona-platform-robot:${ROBOT_TOKEN}@github.com".insteadOf "https://github.com"
+          git config user.name "GitHub Action"
+          git config user.email "github-action@users.noreply.github.com"
+          git config user.password "${ROBOT_TOKEN}"
+          echo "GIT_USER=percona-platform-robot:${ROBOT_TOKEN}" >> $GITHUB_ENV
+
+      - name: Run MkDocs (2.x)
+        if: github.ref == 'refs/heads/main'
+        run: |
+          git branch -D publish || : # delete the branch to start fresh
```
PaulJacobs-percona, line 50: Can use `mike delete --all -b publish -p` but first need to preserve or copy from somewhere `index.html` and `versions.json`

atymchuk: OK, I'll try to make use of `https://github.com/percona-platform/upload-artifact` action

PaulJacobs-percona: I'm forgetting that the branch contains other built version for 1.x and might have more if we extend the list of versions (say, `dev` or `latest`). All we can do is delete this build's directory (`mike delete 2.x -b publish -p`) but I'm not sure it saves much.

PaulJacobs-percona:
@atymchuk I've manually copied build.yml to main branch, leaving out the delete branch step and replacing it with deleting only the version dir (mike delete). I've also deleted the preview.yml workflow which was a failed experiment.

PR #378, opened 2021-02-08
Spellcheck report

PaulJacobs-percona:


atymchuk (approved):
I leveraged an action to install node v14, otherwise it would be node v8, which is too old

PR #382, opened 2021-02-09
PMM-7306 Add description of disable collectors flag.

rnovikovP:


rnovikovP:
@askomorokhov  can we also add some examples for every Service and have a URL to Exporter pages, so the user is able to see other possible options for this feature?

PaulJacobs-percona (approved):

PR #388, opened 2021-02-10
Separate jobs

PaulJacobs-percona:
Separated into 3 dependent stages

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,6 @@ jobs:
   setup:
     runs-on: ubuntu-latest
     steps:
-
       - name: Check out code
         uses: percona-platform/checkout@v2
         with:
@@ -35,7 +34,7 @@ jobs:
           sudo chmod +x /usr/local/bin/plantuml
 
       - name: Install Node.js 14.x
-        uses: percona-platform/setup-node@v2.1.2
+        uses: percona-platform/setup-node@v2
         with:
           node-version: "14"
 
@@ -53,7 +52,6 @@ jobs:
     needs: setup
     runs-on: ubuntu-latest
     steps:
-
       - name: Run MkDocs (2.x)
         if: github.ref == 'refs/heads/main'
         run: |
@@ -64,7 +62,6 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
     steps:
-
       - name: Spelling
         run: |
           sudo npm i markdown-spellcheck -g
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,5 +73,5 @@ jobs:
       - name: Grammar
         run: |
           sudo npm i write-good -g
-          write-good docs/**/*.md || true
+          write-good --no-passive docs/**/*.md || true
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -3,15 +3,15 @@ name: BUILD PMM DOCS (2.x)
 on:
   push:
     branches:
-      - main
+      - separate-jobs
```
atymchuk, line 6: Is this branch long-lived? Cause if you delete it, the workflow will never run.

.github/workflows/build.yml
```diff
@@ -34,6 +34,11 @@ jobs:
           sudo cp $GITHUB_WORKSPACE/_resources/bin/plantuml /usr/local/bin
           sudo chmod +x /usr/local/bin/plantuml
 
+      - name: Install Node.js 14.x
+        uses: percona-platform/setup-node@v2.1.2
```
atymchuk, line 38:
```suggestion
        uses: percona-platform/setup-node@v2
```
We upgraded the script, it should now be just v2.

.github/workflows/build.yml
```diff
@@ -44,16 +49,21 @@ jobs:
           git config user.password "${ROBOT_TOKEN}"
           echo "GIT_USER=percona-platform-robot:${ROBOT_TOKEN}" >> $GITHUB_ENV
 
+  build:
```
atymchuk, line 52: Not sure we need this as a separate step. We could continue running steps in the same job. Whenever a step fails, the whole job fails. Unless you want to continue despite a failing job, this is probably not necessary.

.github/workflows/build.yml
```diff
@@ -44,16 +49,21 @@ jobs:
...
         if: github.ref == 'refs/heads/main'
         run: |
           mike deploy 2.x -b publish -p
           mike retitle 2.x "PMM2 (LATEST)" -b publish -p
 
-      - name: Install Node.js 14.x
-        uses: percona-platform/setup-node@v2.1.2
-        with:
-          node-version: "14"
+  check:
```
atymchuk, line 63: Same as above.

atymchuk (changes requested):

.github/workflows/build.yml
PaulJacobs-percona: Right, it was just for testing, I intended fixing it after merge

.github/workflows/build.yml
PaulJacobs-percona: The whole point of this PR was to separate into jobs. There was no other change. The reason is that you can't see what stage has failed without drilling down. With jobs, you see immediately which part of setup/build/(spell/grammar) check has a problem.

atymchuk: It's an overkill, since every job starts a fresh operating system, which slows down the workflow and burns the resources. 

You can see what stage has failed without drilling down. Example attached.

Screenshot 2021-02-12 at 12 06 24

But I leave it up to you :)

atymchuk (approved):

PaulJacobs-percona:
Convinced not a good way to see separate stage statuses.

PR #410, opened 2021-03-08
RHEL/CentOS 6 only has support up to version 2.12

david-ducos-percona:
Changing the doc to show the real version supporting for RHEL/CentOS 6


david-ducos-percona:
I don't expect to keep the same word, I created this PR to keep up the real world.

PaulJacobs-percona:
Thanks. Already fixed with Roma's PR.

PR #411, opened 2021-03-09
User names clarification

rnovikovP:
There are different types of users in VM for PMMM. one for SSH and another for web UI and adding Services to monitoring

### DIFF
--- a/docs/setting-up/server/virtual-appliance.md
+++ b/docs/setting-up/server/virtual-appliance.md
@@ -27,8 +27,8 @@ Most steps can be done with either a user interface or on the command line, but 
 			- Disk 1 (sda): VMDK (SCSI, 40 GB)
 			- Disk 2 (sdb): VMDK (SCSI, 400 GB)
 	- Username/default password:
-		- `root`/`percona`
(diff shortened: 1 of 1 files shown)


docs/setting-up/server/virtual-appliance.md
```diff
@@ -27,8 +27,8 @@ Most steps can be done with either a user interface or on the command line, but
 			- Disk 1 (sda): VMDK (SCSI, 40 GB)
 			- Disk 2 (sdb): VMDK (SCSI, 400 GB)
 	- Username/default password:
-		- `root`/`percona`
-    	- `admin`/`admin`
+		- `root`/`percona` (SSH)
+    	- `admin`/`admin` (PMM server account )
```
PaulJacobs-percona, line 31: Thanks. But admin/admin is also a CentOS Linux user account, and the user can connect with any account's SSH so long as they set up their keys. This clarification box is for the VM as a Linux instance.

PaulJacobs-percona: FYI The PMM User admin/admin is mentioned in section 5.

PR #435, opened 2021-04-01
PMM-6837 PMM Server and Client with docker-compose

PaulJacobs-percona:


PaulJacobs-percona:
Reviewers: Tested on Debian 10 and RHEL8 but without adding clients, only checking containers run and register.

paskal:
Thanks for inviting me, I'll try to check by the end of the day.

paskal (commented):
LGTM except for the fact I wrote instruction for generic docker-compose (v2, more on differences with v3 [there](https://stackoverflow.com/a/53636006/961092) for example) and the instruction here is written for swarm mode, which no one I know use.

If you would ask me, doing the instruction for docker-compose without swarm and without explicit network configuration (leaving the defaults) would be way easier, and if someone would want to build something more complicated on top of it they would be able to do it themselves.

If the instruction stays in the current state, I would propose to check which features of [v3 docker-compose file](https://docs.docker.com/compose/compose-file/compose-file-v3/) you are using and sticking to the minimum version with all of them, and not to the most recent one as it restricts usage in the older system like almost any LTS Linux: it's a complication for users for no benefit to the information which is delivered by the documentation.

paskal:
I've checked and I see no new features used, so either '3' or even '2' could be used likely without file alterations, as you don't use the `deploy` directive, which is the only one which [was introduced in 3.0](https://docs.docker.com/compose/compose-file/compose-versioning/).

PaulJacobs-percona:
Thanks @paskal
I used swarm as it seemed to be the only way to refer to a separate Docker container by name rather than IP. I regarded the pain of setting up the swarm to be worth the benefit of having example code that 'just works' (no need to say 'use your server IP here').
The version (3.6) was the highest that works for me on all 3 flavors tried (Debian 10, RHEL 8, & Ubuntu 20.10). I will look into finding the lowest possible supported version that coincides with the platforms we officially support (Deb9/10, RH 7/8, Ubuntu 16/18/20). I will also do another batch of research using version 2 and post that compose file, but it will come in another period of focus on this subject.
Thanks again for your insights and for the initial compose examples and tips.

PR #445, opened 2021-04-08
PMM-7816 Add a notice about service parameters changes

adivinho:


PaulJacobs-percona:
I'd prefer mentioning this general tip once, here:
https://www.percona.com/doc/percona-monitoring-and-management/2.x/setting-up/client/index.html#configure-add-services
Later I will expand this section with more general advice and examples of adding services. The reader then dives down into specific sections which are uncluttered with repetitions of general advice. What do you think?

PaulJacobs-percona (approved):

PR #485, opened 2021-05-10
PMM Client: fix creating MongoDB role

maxkondr:


PaulJacobs-percona (approved):

akira-kurogane:
@maxkondr 

The error _"Roles on the 'test' database cannot be granted privileges that target other databases or the cluster"_ is happening because this createRole command is not being run in the "admin" database; the connection is in the default "test" db namespace instead. A common gotcha.

Change to `db.getSiblingDB("admin").createRole({ .. ` instead.

akira-kurogane:
.... and I see you've already done.

Ignore the above. When I landed on the page I thought the context was the error still existed.

PR #493, opened 2021-05-13
PMM-7934 Backup/restore

PaulJacobs-percona:
Notes:
- Section moved from 'How to' to 'Using' as IMHO it is more of a feature than a task, but open to discussion.
- Screenshots & icons need adding/updating, pending a more complete working env.
- Preview: https://609d0fc79be80e00078e07ee--pmm-doc-preview.netlify.app/using/backup.html


docs/using/backup.md
```diff
@@ -0,0 +1,98 @@
+# Backup and Restore
+
+> <b style="color:goldenrod">Caution</b> Backup and restore features are a technical preview and currently only work with MySQL database servers backing up to Amazon AWS S3 storage. locations
```
atymchuk, line 3:
```suggestion
> <b style="color:goldenrod">Caution</b> Backup and restore features are a technical preview and currently only work with MySQL database servers backing up to Amazon AWS S3 storage locations.
```

docs/using/backup.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Before you start
+
+**For restoring**
+
+There are no special requirements for restoring.
```
atymchuk, line 9: Aren't the requirements same as for backing up? i.e., you gotta have access to S3, connectivity, the service has be there, etc. Well, probably a bit different, but still...

PaulJacobs-percona: Yes I realise I misinterpreted some facts. Restoring can be done without installing xtrabackup et al, but sure, still needs the BU location and a server to put it. I'll amend that.

docs/using/backup.md
```diff
@@ -0,0 +1,93 @@
...
+1. Navigate to Backup Management-->Storage locations
+2. Click *Add*
+3. Fill in the form fields
+
+    ![!](../_images/PMM_Backup_Management_Locations_Add_Storage_Location.jpg)
+
+    - *Name*:
+    - *Description*:
+    - *Type:*
+        - *S3*:
```
oter, line 43: Probably we should mention that `S3` is only implemented for now, `Local Client` and `Local Server` are not.

docs/using/backup.md
```diff
@@ -0,0 +1,93 @@
+# Backup and Restore
+
+> <b style="color:goldenrod">Caution</b> Backup and restore features are a technical preview and currently only work with MySQL database servers backing up to Amazon AWS S3 storage locations.
+
+## Before you start
```
oter, line 5: pmm-agent should be run as root in order to stop and restart MySQL using systemctl.

docs/using/backup.md
```diff
@@ -0,0 +1,93 @@
...
+    - *Location* -- The predefined storage location
+
+4. Click *Backup*
+
+5. In the *Backup Inventory* pane, watch the *Status* column.
+
+
+## Restoring a backup {: #backup-restore }
+
+> You can only restore a backup to the same service type. I.e. a MySQL backup can only be restored to a MySQL database server.
```
oter, line 76: restoring works for the same service, e.g. we cannot restore a backup made from `mysql-service-1` to some other `mysql-service-2`. Later, we will add support for restoring to another version-compatible service.

docs/using/backup.md
```diff
@@ -0,0 +1,93 @@
...
+> You must add a [backup storage location](#backup-location) before backing up
+
+1. Navigate to *Backup Management-->Backup Inventory*
+
+2. Click {{icon.plussquare}} Add
+
+3. In the *Backup On Demand* dialog, enter values for:
+
+    - *Service name* -- The service being backed up
+    - *Vendor* -- The database vendor name (only * * currently supported)
```
Dasio, line 64: vendor is read only. User don't have to enter values. And for now only mysql is supported.

docs/using/backup.md
PaulJacobs-percona: `pmm-admin config` starts `pmm-agent` so if that command is sudo'd, won't pmm-agent also be root PID? But you raise a general point as to whether all commands should show they are run as `sudo` or should be assume the reader can work that out? In the case of docker containers, `sudo` fails because it's not in the provided image. I will defer this specific instance and do a global doc review of where/when/if to mention running all `pmm-admin` commands via sudo.

docs/using/backup.md
PaulJacobs-percona: So not only same service type, but you can only restore to the same service instance (name)?

docs/using/backup.md
oter: As far as I know when we run `pmm-admin config` it actually performs a configuration (modifies configuration file pmm-agent.yaml) and this command requires pmm-agent to be already running via `pmm-agent run --server-address=1.2.3.4:443 --config-file=pmm-agent.yaml`. One more thing, it is required that the running agent should be on the same machine as MySQL e.g. the restore won't work if pmm-agent inside Docker container and MySQL is on the host machine.

docs/using/backup.md
oter: yes

PaulJacobs-percona: Ok I will add some words saying 'In the current implementation, you can only restore... More flexibility will come in future releases.' or something like that.

docs/using/backup.md
PaulJacobs-percona: Thanks @oter for the clarifications. These are general pointers that we will cover in other sections. I have a place in mind and will tackle it separately. If you think it's important to give these tips on this page (right next to the commands), let me know.

PR #524, opened 2021-06-16
Simplify the docker commands

atymchuk:
Simplified two things:
1/ this script does run without the ln command, so it is redundant
2/ no need to output the health status in raw format, it works fine without it


PaulJacobs-percona:
The `{% raw %}` / `{% endraw %}` are needed to prevent MkDocs errors when it tries to interpret Jinja-like variables as used by docker. They are not seen in the final output and should remain in place.
For anything else I assume the code works as this piece was originally inserted by PZ.

PR #528, opened 2021-06-17
PMM Server Docker vs VM vs AWS: pros/cons

PaulJacobs-percona:


docs/setting-up/index.md
```diff
@@ -13,9 +13,11 @@ Set up at least one PMM Server.
...
 Choose from:
 
-- [Docker](server/docker.md)
-- [Virtual appliance](server/virtual-appliance.md)
-- [Amazon AWS](server/aws.md)
+| Use                  | {{icon.thumbsup}} **Benefits**  | {{icon.thumbsdown}} **Drawbacks**
+|----------------------|---------------------------------|--------------------------------------
+| [Docker]             | Quick, simple                   | Docker required, not for production
+| [Virtual appliance]  |                                 |
+| [Amazon AWS]         |                                 | Non-free solution (infrastructure)
```
PaulJacobs-percona, line 21: Fill in the blanks!

PaulJacobs-percona:
Idea is to bring some guidance to the options presented in the 'Setting up' sections.
Starting with pros/cons for each of three (maybe 4 if we include the install one-liner).
Please fill in the blanks.

docs/setting-up/index.md
```diff
@@ -13,9 +13,11 @@ Set up at least one PMM Server.
 
 Choose from:
 
-- [Docker](server/docker.md)
-- [Virtual appliance](server/virtual-appliance.md)
-- [Amazon AWS](server/aws.md)
+| Use                  | {{icon.thumbsup}} **Benefits**  | {{icon.thumbsdown}} **Drawbacks**
+|----------------------|---------------------------------|--------------------------------------
+| [Docker]             | Quick, simple                   | Docker required, not for production
```
shoffman-percona, line 18:
```suggestion
| [Docker]             | Quick, simple                   | Docker required, will have additional network configuration needs
```

docs/setting-up/index.md
```diff
@@ -13,9 +13,11 @@ Set up at least one PMM Server.
...
+| [Virtual appliance]  |                                 |
```
shoffman-percona, line 19:
```suggestion
| [Virtual appliance]  | Easily import into Hypervisor of your choice  | Requires more system resources compared to docker footprint
```

docs/setting-up/index.md
```diff
@@ -13,9 +13,11 @@ Set up at least one PMM Server.
...
+| [Amazon AWS]         |                                 | Non-free solution (infrastructure)
```
shoffman-percona, line 20:
```suggestion
| [Amazon AWS]         |  Wizard driven install                               | Non-free solution (infrastructure)
```

shoffman-percona (commented):
I can add more, am I on the right track to what you were looking for?

PaulJacobs-percona:
> I can add more, am I on the right track to what you were looking for?

Add as many as you like. I will 'bulletise' them is possible. Idea is for reader to match their use case to a choice of options.

In many places in the doc we say 'here are the options' or 'this step is optional' but don't explain the consequences or give our advice.

docs/setting-up/index.md
```diff
@@ -13,9 +13,11 @@ Set up at least one PMM Server.
 
 Choose from:
 
-- [Docker](server/docker.md)
-- [Virtual appliance](server/virtual-appliance.md)
-- [Amazon AWS](server/aws.md)
+| Use                  | {{icon.thumbsup}} **Benefits**  | {{icon.thumbsdown}} **Drawbacks**
+|----------------------|---------------------------------|--------------------------------------
+| [Docker]             | Quick, simple                   | Docker required, will have additional network configuration needs
```
nikita-b, line 18: What network configurations are we referring to?
We have additional network layer with Docker but:
1. This usually doesn't affect user experience.
2. We can use net=host and it'll be same as virtual machine. (From network side)

My opinion:
- Benefits: Quick and simple start, portability (Easy to move instance to another server or datacenter), Requires less system resources compared to Virtual appliance
- Drawbacks: Docker is additional level of abstraction + Need to have (and maintain) server (Or virtual machine) with Docker.

shoffman-percona: our instructions don't mention net=host anywhere...so if users follow our instructions they tend to get stuck with the port forwarding approach and then tangled in iptables/firewalld issues especially when testing pmm out with both a db server (on host) and pmm-server (in container)...can remove was just capturing thoughts.  

I think there is a drawback of docker being more ephemeral vs installing a bunch of components on a server...lots of people get bit by modifying /etc/grafana/X and then upgrade only to realize their config changes are lost...

nikita-b: Did we have a lot of tickets with Docker network-related issues? I just haven't seen such problems for a long time, but maybe this is just a `blind spot` for me.

But I totally agree about configs/upgrade issue :(

Maybe we can add /srv/grafana/custom.ini for Grafana and add a note: "DON'T CHANGE IT" to the main config file. (And PMM docs).

PR #533, opened 2021-06-22
Small improvement for mongodb agent docs

nikita-b:
1. Added warning about `rateLimit` in a config file block.
2. Moved tips before examples. I think it's more natural because if you see examples you want to run it. 
3. Added link to `pmm-admin add mongodb`options.

### DIFF
--- a/docs/setting-up/client/mongodb.md
+++ b/docs/setting-up/client/mongodb.md
@@ -105,7 +105,7 @@ You can set profiling:
     operationProfiling:
       mode: all
       slowOpThresholdMs: 200
-      rateLimit: 100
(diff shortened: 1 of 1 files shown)


PaulJacobs-percona (approved):
Thanks @nikita-b

PaulJacobs-percona:
@nikita-b I subsequently removed the link to pmm-admin in the Tips box as there is already one at the bottom of the page in the 'see also' box. Yes it is away from the action but I prefer putting them at then end rather than scattered throughout the content. It avoids distraction.

nikita-b:
@PaulJacobs-percona ok, it makes sense. It was unclear for me what `--cluster` means but we have a description in the tips box. The rest of the options seem obvious.

PR #547, opened 2021-07-07
PMM-8239 PMM2 does not add Azure database with ssl

adivinho:


denisok (approved):

PaulJacobs-percona (approved):
There's a few changes I'd like to make but it's okay as is and I'd be happier revising the whole page in context with this new inclusion.

PR #559, opened 2021-07-21
Feedback from forum user

shoffman-percona:
A user on the forums interpreted this as a command they must run on a newly created PMM server from AWS Marketplace and broke their install badly!  Needed to clarify that is an example only for another DB host and should NOT be run on the PMM server.  Feel free to reword!

### DIFF
--- a/docs/setting-up/server/aws.md
+++ b/docs/setting-up/server/aws.md
@@ -111,7 +111,7 @@ You are creating a username and password that will be used for two purposes:
 
 1. authentication as a user to PMM - the credentials to log in to PMM.
 
-2. authentication between PMM Server and PMM Clients - you will re-use these credentials when configuring PMM Client for the first time on a server, for example:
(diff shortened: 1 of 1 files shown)


rnovikovP:
I think all the "Installation steps" should be done on https://github.com/percona/pmm-doc/blob/main/docs/_images/PMM_Login.jpg this image. 

Everything below (about user name ) is redundant as this is the same information we would tell for docker, ovf, or other installations. 

So ur Server installation instructions must be ended in this stage.

my proposal - remove all sections about admin/admin and pmm-configure. but make sure this is explained somewhere.

Note: we are NOT expecting users SSH much to PMM or go inside docker .. this all "workaround cases"  for troubleshooting section

PaulJacobs-percona (approved):
AWS page is due a wholesale rework so wording/logic/admonitions will be further improved.

PR #561, opened 2021-07-22
pmm-admin command fails with unknown flag

riveraja:
The flags --SERVER_USER and --SERVER_PASSWORD (in uppercase letters) aren't accepted by the pmm-admin command in the last version of pmm-client v1. Although PMMv1 is not actively developed anymore in favor of PMMv2 at least make the documentation match with the CLI command.


PaulJacobs-percona (approved):
Many thanks!

PR #568, opened 2021-08-03
Updating instructions for 2.20.0 RN

shoffman-percona:
Found the blurb about the issue starting in 2.16.0 to be confusing so tried to clarify it a bit more.  Since the two commands get the user back up and running I moved that up over a backup and restore since we gloss over that users may have installed their original image with a complex docker run command.


ggaspard (approved):
Looks good to me.

rnovikovP (approved):

PaulJacobs-percona (approved):

PR #570, opened 2021-08-04
Backup scheduling:  add retention field

fabio-silva:
Update docs with retention for backup scheduling.


docs/using/backup.md
```diff
@@ -161,9 +161,11 @@ Make regular scheduled backups.
             - *Start time, h/m*: The hour and minute for the backup.
                 - In the first field, select one or more hours (*00* to *23*, *00* is midnight).
                 - In the second field, select one or more minutes (*00* to *59*).
+
+        - *Retention*: Number of backups to keep. Insert 0 if there is no limit.
```
PaulJacobs-percona, line 165:
```suggestion
        - *Retention*: How many backups to keep. For unlimited, use `0` (zero).
```

PaulJacobs-percona: What happens when space is limited but backups are unlimited? Is there a warning?

docs/using/backup.md
```diff
@@ -161,9 +161,11 @@ Make regular scheduled backups.
...
         - *Enable*: Deselect to define the scheduled backup without enabling it.
 
-        ![!](../_images/PMM_Backup_Management_Schedule.png)
+        ![Screenshot 2021-08-04 at 11 33 23](https://user-images.githubusercontent.com/4190654/128168942-d2d02cf7-e885-424b-8033-84499a2b98e4.png)
```
PaulJacobs-percona, line 168: Please put the image locally (in docs/_images) so that it can be refreshed or replaced in the future.

PaulJacobs-percona (changes requested):
Please see inline comments.

docs/using/backup.md
fabio-silva: @Dasio  any thought about this?☝️

Dasio: No, we don't check storage. We don't know actual backup size before it is made. So it there will be not enough space, then it will fail.

PaulJacobs-percona (approved):

PR #573, opened 2021-08-10
PMM-5669 Fix for hardcoded paths in PMM.

JiriCtvrtka:


PaulJacobs-percona (approved):

docs/details/commands/pmm-agent.md
```diff
@@ -46,6 +46,7 @@ Most options can be set via environment variables (shown in parentheses).
 | `--machine-id=machine-id`              | `PMM_AGENT_SETUP_MACHINE_ID`        | Node machine ID (default is auto-detected).
 | `--metrics-mode=auto`                  | `PMM_AGENT_SETUP_METRICS_MODE`      | Metrics flow mode for agents node-exporter. Can be `push` (agent will push metrics), `pull` (server scrapes metrics from agent) or `auto` (chosen by server).
 | `--node-model=NODE-MODEL`              | `PMM_AGENT_SETUP_NODE_MODEL`        | Node model.
+| `--paths-base=PATH`                    | `PMM_AGENT_PATHS_BASE`              | Base path for PMM client.
```
denisok, line 49: could you please be more explicit here and define what it is, that is is actually path to all binaries ? Because ppl could read it as BASE for all other PATHs that would be relative

denisok (changes requested):

docs/details/commands/pmm-agent.md
JiriCtvrtka: Added. @PaulJacobs-percona please can you check description of flag here: https://github.com/percona/pmm-doc/pull/573/files#diff-ff428c96162b07ed27ea02832cbea69ae4aa4dbedd861426bd8a82c9535d24f2R68 and make it more clear for customers? Thanks.

docs/details/commands/pmm-agent.md
```diff
@@ -64,6 +65,13 @@ Most options can be set via environment variables (shown in parentheses).
 | `-h`, `--help`                         |                                     | Show help (synonym for `pmm-agent help`).
 | `--version`                            |                                     | Show application version, PMM version, time-stamp, git commit hash and branch.
 
+Flag --paths-base is a directory where all binaries, tools and collectors are located.
```
PaulJacobs-percona, line 68:
```suggestion
- The value for `--paths-base` is the path to the directory where all binaries, tools and collectors are located.
```

denisok (approved):

docs/details/commands/pmm-agent.md
```diff
@@ -46,7 +46,8 @@ Most options can be set via environment variables (shown in parentheses).
 | `--machine-id=machine-id`              | `PMM_AGENT_SETUP_MACHINE_ID`        | Node machine ID (default is auto-detected).
 | `--metrics-mode=auto`                  | `PMM_AGENT_SETUP_METRICS_MODE`      | Metrics flow mode for agents node-exporter. Can be `push` (agent will push metrics), `pull` (server scrapes metrics from agent) or `auto` (chosen by server).
 | `--node-model=NODE-MODEL`              | `PMM_AGENT_SETUP_NODE_MODEL`        | Node model.
-| `--paths-exporters_base=PATH`          | `PMM_AGENT_PATHS_EXPORTERS_BASE`    | Base path for exporters to use.
+| `--paths-base=PATH`                    | `PMM_AGENT_PATHS_BASE`              | Base path for PMM client, where all binaries, tools and collectors are located. If not set, default is `/usr/local/percona/pmm2`.
```
PaulJacobs-percona, line 49: @denisok @JiriCtvrtka I rewrote the path comments to be inline, within the table. For even more clarification, we can add some examples where paths-base/paths-exporters_base is/is not given, and the resulting paths for each.

One other comment. It's probably asking for trouble to use mixed dash and underscore within a parameter name. They are easy to mistype or misread. Is it too late to standardise on full dashes, as all the other option names seem to be?

docs/details/commands/pmm-agent.md
```diff
@@ -46,7 +46,8 @@ Most options can be set via environment variables (shown in parentheses).
...
+| `--paths-exporters_base=PATH`          | `PMM_AGENT_PATHS_EXPORTERS_BASE`    | Base path for exporters to use. If not set, or set to a relative path, uses value of `--paths-base`.
```
PaulJacobs-percona, line 50:
```suggestion
| `--paths-exporters_base=PATH`          | `PMM_AGENT_PATHS_EXPORTERS_BASE`    | Base path for exporters to use. If not set, or set to a relative path, uses value of `--paths-base` prepended to it.
```

docs/details/commands/pmm-agent.md
denisok: unfortunately those are old ones, we can't rename it, ppl probably already use them in production. We can add new and have them in documentation and leave old ones in the code, so both works. But it is too late now.

docs/details/commands/pmm-agent.md
denisok: it is true for all other exporters and tools ("--paths-pt-summary=" and etc), so we need to add to all of them.

docs/details/commands/pmm-agent.md
JiriCtvrtka: Basically if you pass absolute path to any param then it will be respected, othervise it will be based on paths-base param. Expect case when you pass exporter base and paths base together. Paths base will be used every where, but for exporters will be respected exporters base value.

docs/details/commands/pmm-agent.md
JiriCtvrtka: Paths base will applied to everything (tools, exporter, collectors). If you will pass for example node exporter path which is absolute then this value will be respected othervise if it is relative it will be based on paths-base value. Only one exception is when you passed paths-base and expoter-base together. Is exporter base is absolute then it will be applied on every exporter path other things will has path based on paths base value.

PaulJacobs-percona:
@JiriCtvrtka What's the status of this? Still pending/alive?

JiriCtvrtka:
@PaulJacobs-percona Yes it is still alive. This task waiting for QA, already 16 days due to long QA queu.

PaulJacobs-percona:
Pending QA testing

PR #577, opened 2021-08-13
clarify no downgrades

shoffman-percona:
since htis isn't called out...we say you can't downgrade internally..probably should tell the world!

### DIFF
--- a/docs/setting-up/server/docker.md
+++ b/docs/setting-up/server/docker.md
@@ -101,6 +101,9 @@ How to run PMM Server with Docker based on our [Docker image].
 
 ---
 
+!!! caution alert alert-warning "Important"
+    Downgrades are *NOT* supported.  In order to revert to a previous version you will need to ensure you take a backup of that version before you do the initial upgrade. 
+
 !!! hint alert alert-success "Tip"
     To see what release you are running, use the *PMM Upgrade* panel on the *Home Dashboard*, or run:
 
@@ -130,6 +133,7 @@ How to run PMM Server with Docker based on our [Docker image].
     --name pmm-server \
     percona/pmm-server:2
     ```
+4. Perform a [restore](#restore)
 
 ## Restore
 


docs/setting-up/server/docker.md
```diff
@@ -101,6 +101,9 @@ How to run PMM Server with Docker based on our [Docker image].
 
 ---
 
+!!! caution alert alert-warning "Important"
```
ggaspard, line 104: The backup is just for PMM, right? No databases are affected?

docs/setting-up/server/docker.md
```diff
@@ -101,6 +101,9 @@ How to run PMM Server with Docker based on our [Docker image].
...
+    Downgrades are *NOT* supported.  In order to revert to a previous version you will need to ensure you take a backup of that version before you do the initial upgrade.
```
PaulJacobs-percona, line 105:
```suggestion
    Downgrades are not possible. To go back to using a previous version you must have created a backup of it before upgrading.
```

PaulJacobs-percona: From my reading this admonition has two conflicting messages. First 'you can't downgrade', second 'you can IF you made a backup of whatever you want to downgrade to before upgrading'.
1. Is that a true interpretation?
2. Can we mention why downgrading is not technically possible.
3. And then what the workaround is, and what you loose (I'm guessing the data collected since the downgraded version).
Also I would prefer this aside be moved to the top of this 'upgrade' section, so readers are fully aware before they start reading the steps. It also keeps the steps themselves clean and clear of clutter. An alternative (a bit mean but perhaps better) is to have a new section 'Downgrade' where we say 'you can't unless you've done X, Y and when you do you will suffer Z, so here are the precise steps for that'.

rnovikovP: There is no Automatic downgrade.. so the BackUp is the only  way
And yes - you'll lose the data collected while you were on a newer version or instances added during this period of time

shoffman-percona: you're not actually downgrading...you are restoring to an old version and point in time...as if the upgrade never took place.  We can mention all kinds of good stuff to the user but does that go here or it's own sections on downgrades.  My main concern is that it be called out VERY clearly at the start of the upgrade process, how best to do that I leave to you all!

PaulJacobs-percona: Agree that any mention of 'downgrade' here is deceptive. Perhaps a FAQ entry instead? "How do I downgrade?" with
> You can't. But you can take a backup of a version before upgrading and revert back to it, but you will loose any changes made since.

docs/setting-up/server/docker.md
PaulJacobs-percona: Yes @ggaspard

PR #585, opened 2021-08-30
Common Linux/macOS command for shasum

PaulJacobs-percona:
macOS 11.5 and Debian 11 both allow `shasum -a` option. Command change to use this approach to make a common code block.
https://deploy-preview-585--pmm-doc.netlify.app

### DIFF
--- a/docs/setting-up/server/easy-install.md
+++ b/docs/setting-up/server/easy-install.md
@@ -3,22 +3,12 @@
 !!! caution alert alert-warning "Caution"
     Download and check `get-pmm.sh` before running it to make sure you know what it does.
 
-## Linux
+## Linux or macOS
 
 ```sh
 curl -fsSL -O https://raw.githubusercontent.com/percona/pmm/PMM-2.0/get-pmm.sh \
 -O https://raw.githubusercontent.com/percona/pmm/PMM-2.0/.sha256-oneline && \
-sha256sum .sha256-oneline -c && \
-chmod +x ./get-pmm.sh && \
-./get-pmm.sh
-```
-
-## macOS
-
-```sh
-curl -fsSL -O https://raw.githubusercontent.com/percona/pmm/PMM-2.0/get-pmm.sh \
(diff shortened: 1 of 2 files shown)


docs/setting-up/server/easy-install.md
```diff
@@ -3,22 +3,12 @@
 !!! caution alert alert-warning "Caution"
     Download and check `get-pmm.sh` before running it to make sure you know what it does.
 
-## Linux
+## Linux or macOS
```
PaulJacobs-percona, line 7: Am I right on Linux this should be run as su? (And that sudo the whole command doesn't work?)

shoffman-percona: I'm lacking context on the question but here's the main difference between the two:
`su` becomes root and then N number of commands are able to be executed (the - adds more power/meaning but not the point here)
`sudo` simply executes a single command as root...so you'd need to put it in front of every command that requires root privs.  

So if we're talking about the "get-pmm.sh" script...you can run that with either `sudo get-pmm.sh` and be done or you can do `su ; get-pmm.sh; exit` because su will keep you as root when you complete the execution to allow more commands to be run.  Fun fact...if you do su - you will also have to cd back to the directory the script is downloaded in because the - actually causes you to inherit the users profile (thus you will change to your home directory /root/). Does that help?

shoffman-percona (approved):
I just ran the command on my laptop as me and it worked for --help. I dont' recall if trying to run the script to perform the install does privilege escalation but I can try to run it later this afternoon for a complete install.

docs/setting-up/server/easy-install.md
PaulJacobs-percona: Thanks for the masterclass! Context is that when I cut'n'paste into macos as is (no sudo, no su), it works. But for Debian, something fails (possibly a Docker command) even when the whole line is `sudo ...`. Only when I `su -` first then recopied/pasted command did it work. 
I just wondered what the official approach was meant to be.

atymchuk (approved):

PR #592, opened 2021-09-10
Remove SUPER from grants

Tusamarco:
As for discussion the use of super seems to be redundant also in the code we see:
```
if !strings.Contains(grants, "RELOAD") && !strings.Contains(grants, "ALL PRIVILEGES") {
		s.l.Error("RELOAD grant not enabled, cannot rotate slowlog")
		return
	}
```
in https://github.com/percona/pmm-agent/blob/d31783a23cfd8cbe098d7e4839ca11c5cb4d16b9/agents/mysql/slowlog/slowlog.go
So we should be good to remove it and count on `RELOAD` instead


beatahandzelova:
@Tusamarco BACKUP_ADMIN role is also needed for pmm to work so https://github.com/percona/pmm-doc/pull/638/files was merged into docs

PR #593, opened 2021-09-14
remove docker compose way from official Doc

rnovikovP:
While docker-compose is widely used internally, it's not an official and supported way of installing PMM.
We can have the instructions in forum or blogs but not in the official documentation as this is confusing

### DIFF
--- a/docs/setting-up/server/docker.md
+++ b/docs/setting-up/server/docker.md
@@ -230,47 +230,6 @@ How to run PMM Server with Docker based on our [Docker image].
     docker rmi $(docker images | grep "percona/pmm-server" | awk {'print $3'})
     ```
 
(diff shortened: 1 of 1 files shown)


puneet0191 (approved):
let me approve it first, thank you

denisok (approved):

PaulJacobs-percona (approved):
"Not officially supported" is imo not a good reason to remove it.

rnovikovP:
PMM Documentation = "Official PMM Documentation," so everything in our DOC should work, and we are obligated to make it work. 
We CAN use internally whatever we need, and we can share this knowledge in our Blog, but documentation needs to be clean from unsupported tools and solutions, so the people are not confused.

PR #604, opened 2021-09-30
Fixes for create user step for RDS/Aurora, which will fail otherwise …

jobinau:
rds_superuser is separate role, not an attribute
The step mentioned in the current document won't work. It will give syntax error.
Corrected the step.


docs/setting-up/client/postgresql.md
```diff
@@ -31,10 +31,12 @@ We recommend creating a PMM database account that can connect to the `postgres`
     CREATE USER pmm WITH SUPERUSER ENCRYPTED PASSWORD '******';
     ```
 
-    If your database runs on Amazon RDS:
+    If your database runs on Amazon RDS / Aurora PostgreSQL, The SUPERUSER cannot be assigned.  
+    So we have to create the user first and then grant the `rds_superuser` role to it.
 
     ```sql
-    CREATE USER pmm WITH rds_superuser ENCRYPTED PASSWORD '******';
+    CREATE USER pmm WITH PASSWORD '******';
```
denisok, line 38: ENCRYPTED is missed on purpose? we probably still recommend that

rnovikovP: based  on https://jira.percona.com/browse/PMM-9453  , yes we should keep it encrypted

beatahandzelova (approved):

rasika-chivate (approved):

PR #605, opened 2021-09-30
Remove --slow-log-rotation option reference

mlarin05:
The option '--slow-log-rotation' is not supported by pmm-admin 2.x. 
Instead, the negative value of '--size-slow-logs' is used to disable slow log rotation.
This is a fix for the https://jira.percona.com/browse/PMM-8890


denisok (approved):

denisok:
@mlarin05 please sign CLA and we will land!

PR #611, opened 2021-10-06
PMM-8970 Installed custom Grafana plugins lost after upgrade via Docker

adivinho:


denisok:
@adivinho could you also add some note to the ticket (in ADDITIONAL RN or in desc) that this will be fixed automatically only from 2.23.0 forward. Because it is misleading from the ticket desc that is fixes the docker upgrade also.

PR #613, opened 2021-10-11
Clarification on plugins problem

rnovikovP:
Trying to add more clarity on the plugins problem


adivinho:
> @adivinho , can we help people to easily see what plugins they need to re-install? and can they use the code from the currently running instance to use data for restore instead of reinstalling plugins?

IMHO it's much easier to reinstall plugins if a person managed to install additional plugins earlier. 
At least it's only one command instead of whole bunch of checks and copy commands.

Also we have got no reports regarding the issue from users till now. 

I can add a list of plugins that are installed by default but can we guaranty that this list will be keep up-to-date and doesn't bring more confusions in a future.

PR #620, opened 2021-11-02
PMM-9051 Updated docs on custom queries

percona-csalguero:


rnovikovP (approved):
Adding more information is 👍 
However, I'm not clear on the goals we are trying to achieve here, so adding  @denisok  to review too.

denisok:
@rnovikovP document user feedback ? So each user wouldn't need to find it out from scratch again?

denisok (approved):

PR #621, opened 2021-11-09
make update check wording less confusing

shoffman-percona:
Feedback from Ananias that the Disable Updates env flag is confusing so attempting to make it less so.

### DIFF
--- a/docs/setting-up/server/docker.md
+++ b/docs/setting-up/server/docker.md
@@ -246,7 +246,7 @@ Use the following Docker container environment variables (with `-e var=value`) t
 
 | Variable                   | Description
 | -------------------------- | -----------------------------------------------------------------------
-| `DISABLE_UPDATES`          | Disable automatic checks for updated PMM versions. *Updates are not automatically applied*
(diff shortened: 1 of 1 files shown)


docs/setting-up/server/docker.md
```diff
@@ -246,7 +246,7 @@ Use the following Docker container environment variables (with `-e var=value`) t
 
 | Variable                   | Description
 | -------------------------- | -----------------------------------------------------------------------
-| `DISABLE_UPDATES`          | Disable automatic updates.
+| `DISABLE_UPDATES`          | Disable automatic checks for updated PMM versions. *Updates are not automatically applied*
```
atsaloux, line 249: `automatic` is not ideal for me as that could lead to the assumption that checks for updates are still possible -even if not automatic - , which is not the case.

I'd phrase that like `Disable the ability to check for new PMM versions and apply upgrades using the UI`

atymchuk (approved):
👍

rnovikovP (approved):

PR #624, opened 2021-11-10
PMM-8588 Histogram.

JiriCtvrtka:


docs/using/query-analytics.md
```diff
@@ -139,6 +139,24 @@ For PostgreSQL queries (when using *pg_stat_monitor*) the top query will also be
 
 ![!image](../_images/PMM_Query_Analytics_Tabs_Details_TopQuery.png)
 
+Another metric available (when using *pg_stat_monitor*), especially to monitor PostgreSQL Server performance is *histogram*. 
+*Histograms* provide more explicit information about number of queries for fingerprint (queryid). Ranges are from 0 seconds up to 10 seconds.
```
tiagomotasantos, line 143: 100 esconda right? If the value is in ms than it's 100.

tiagomotasantos (changes requested):

docs/using/query-analytics.md
JiriCtvrtka: Yes 100 seconds thanks. I missed one 0.

tiagomotasantos (approved):

PR #628, opened 2021-11-17
Add connection limit to the PMM user.

guriandoro:
This is in line with the MySQL documentation, in which we show how to limit the amount of connections incoming from the PMM user:
https://www.percona.com/doc/percona-monitoring-and-management/2.x/setting-up/client/mysql.html

CREATE USER 'pmm'@'localhost' IDENTIFIED BY 'pass' WITH MAX_USER_CONNECTIONS 10;


denisok (approved):

rnovikovP:
@puneet0191  pls see this change^. Maybe we should add some basic checks for the number of connections to avoid problems, as we should assume not everyone will follow this recommendation  (

rnovikovP (approved):
Thanks for contributing, @guriandoro!

PR #635, opened 2021-11-25
PMM-9157: Add documentation for changing default timezone in Grafana.

anton-bystrov:
Remove default timezone "browser" in dashboards. Add describe how to change default users timezone.


rnovikovP (approved):
👍

PR #637, opened 2021-11-25
PG-234 Note about the order of extensions

nastena1606:
modified:   docs/setting-up/client/postgresql.md


rnovikovP:
Thanks! 
This is an important note to prevent people from getting in trouble with the wrong configuration

rasika-chivate (approved):

PR #650, opened 2021-12-06
Adding documentation to PG advisors v1

elchinoo:
I've added the documentation files for the PG advisors v1


rnovikovP (commented):
LGTM. 
There will be no direct links to the pages for now, only from Checks, so we can modify them later if needed.

rnovikovP (approved):

PR #654, opened 2021-12-08
Connect pmm platform

catalinaadam:
Updated topic content to reflect changes for connecting PMM to Percona Platform


docs/how-to/configure.md
```diff
@@ -196,38 +196,48 @@ Fill both fields and click the *Apply Alertmanager settings* button to proceed.
...
-3. Choose and enter a strong password in the *Password* field.
-4. Select the check box acknowledging our terms of service and privacy policy.
-5. Click *Sign up*.
+If you see an option to create a new organization instead, your Percona Account is not linked to any organization yet. If this is the case, contact your account administrator, or create a new organization for your Percona Account.
 
-A brief message will confirm the creation of your new account and you may now log in with these credentials.
+### Set the public address of your PMM server 
+1. In PMM, go to *Settings > Advanced Settings*.
+2. Enter your address/hostname or click *Get from browser* to enable your browser to automatically detect and populate this field.
+3. Save the changes.
```
yurkovychv, line 224: according to our [feedback document](https://docs.google.com/document/d/1xSrf3r4ojUhpYxVHjLGc29rpzIcaCBfRcKZ-fewVAGg/edit) we should change `Save the changes` to `Apply Changes` or something similar

docs/how-to/configure.md
```diff
@@ -196,38 +196,48 @@ Fill both fields and click the *Apply Alertmanager settings* button to proceed.
...
-    Your Percona Platform account is separate from your PMM User account.
+## Connect PMM to Percona Platform
+To connect your PMM server to Percona Platform:
+
+1. In PMM go to *Settings > Percona Platform* tab:
+    ![!image](../_images/PMM_Settings_Percona_Platform_Login.png)
+2. Fill in the *Connect PMM to Percona Portal* form with the name of your PMM instance and the credentials of your Percona Account:
+3. Click *Connect*.
+
+After connecting, you can also use SSO for PMM server.
```
yurkovychv, line 234: Maybe we should also clarify that the PMM login screen will contain a separate  button `Sign in with Percona Account` ?

docs/how-to/configure.md
Hlmn-Rmr: @yurkovychv We've discussed this. The intention is not to describe the precise label on the button, that label may change at some point and we would have to update documentation. This step No. 3 doesn't say Click on "Save the changes", in that case it would be a mistake, this step simply says Save the changes. Users can easily correlate that to the "Apply Changes" button.

Hlmn-Rmr (changes requested):
@catalinaadam This PR should target the 2.25.0 branch, not main.

docs/how-to/configure.md
yurkovychv: Agree, just decided to double check

docs/how-to/configure.md
rnovikovP: But the current design is not what we want to have. it needs to be  revoked, maybe even for  2.26, so we'll change it  soon

rnovikovP (approved):

rnovikovP:
@Hlmn-Rmr, @catalinaadam  I've tried to change the base branch ..  but this makes even more problems(

docs/how-to/configure.md
yurkovychv: I'm just saying that we can point a user how and where he/she can use SSO login

denisok:
@catalinaadam it is better to create new branch out of 2.25.0 and then cherry-pick your commits, then create new PR our of the branch based on 2.25.0

catalinaadam:
Recreated this pull request https://github.com/percona/pmm-doc/pull/657

PR #659, opened 2021-12-10
PMM-9298 AMI 8GB

denisok:


rnovikovP:
Can we point people to our instructions on how to increase disk size?

denisok:
> Can we point people to our instructions on how to increase disk size?

@nikita-b would do it

nikita-b:
@denisok @rnovikovP I added instructions to change the disk size but I used Ansible playbooks (It should be easier for users than 10+ command and it's consistent with our normal process but we need to store ansible playbooks somewhere.
I used gist but it looks bad my suggestion:
1. We can store it in pmm-update repo, like: https://raw.githubusercontent.com/percona/pmm-update/main/fix-lvm.yml
2. We can store it in pmm-doc repo, like: https://raw.githubusercontent.com/percona/pmm-doc/main/docs/examples/fix-lvm.yml

rnovikovP:
I think (1) is ok if this is the same set of commands we use in the update.

rnovikovP (approved):

PR #662, opened 2021-12-15
PMM-8867 How to update pmm2-client

anton-bystrov:
Add instruction how to update PMM2-Agent without rewriting config file

### DIFF
--- a/docs/how-to/upgrade.md
+++ b/docs/how-to/upgrade.md
@@ -26,6 +26,16 @@ If one is available, click the update button to update to the version indicated.
 !!! seealso alert alert-info "See also"
     [PMM Server Docker upgrade](../setting-up/server/docker.md#upgrade)
 
+## Updating a PMM-Agent
+
+PMM-Agent can be updated from tarball:
+
+- download tar.gz with pmm2-client
+- unpack it
+- run ./install_tarball script with "-u" flag.
+
+If you want update with rewriting config files run this script without any flags.
+
 ## Upgrade from PMM 1
 
 Because of the significant architectural changes between PMM1 and PMM2, there is no direct upgrade path. The approach to making the switch from PMM version 1 to 2 is a gradual transition, outlined [in this blog post](https://www.percona.com/blog/2019/11/27/running-pmm1-and-pmm2-clients-on-the-same-host/).
--- a/docs/how-to/upgrade.md
+++ b/docs/how-to/upgrade.md
@@ -30,11 +30,11 @@ If one is available, click the update button to update to the version indicated.
 
 PMM-Agent can be updated from tarball:
 
-- download tar.gz with pmm2-client
-- unpack it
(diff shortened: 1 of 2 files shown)


denisok:
`install_tarball` also mentioned here: https://github.com/percona/pmm-doc/blob/aec89f2a9c8b5c657bd52c3dc2679738e5945fd2/docs/setting-up/client/index.md

do we need to update it as well to make sure ppl wouldn't lose their configs if they use that instruction for the update as well?

docs/how-to/upgrade.md
```diff
@@ -26,6 +26,16 @@ If one is available, click the update button to update to the version indicated.
...
+## Updating a PMM-Agent
+
+PMM-Agent can be updated from tarball:
+
+- download tar.gz with pmm2-client
+- unpack it
+- run ./install_tarball script with "-u" flag.
+
+If you want update with rewriting config files run this script without any flags.
+
```
rasika-chivate, line 38: Please find some minor changes:

PPMM-Agent can be updated from tarball:
		
PMM-Agent can be updated from tarball:
		
1. Download tar.gz with pmm2-client.
2. Extract it.
3. Run ./install_tarball script with the "-u" flag.

If you want to update the PMM-Agent with rewriting the config files, run this script without the "u" flag.

rnovikovP: The term "rewrite config file" will not be apparent to the user. Can we explain it more, so users know what and how this will affect them? What will be the difference between with flag and without a flag?

rnovikovP:
⚠️  the bigger problem is: 
As for now, our Doc will have instructions on how to update  Binary, but we don't have "update pmm client by packages" (our primary method).

I see we have https://www.percona.com/doc/percona-monitoring-and-management/2.x/setting-up/client/index.html#binary-package and the same for binary, so we can reference them if needed. But we must have some short instructions for the  update by packages near the binaries update

docs/how-to/upgrade.md
rasika-chivate: I agree. 

We can have this as a hint or note:

Hint! The configuration file will be overwritten if you provide the "-u" flag while the PMM-Agent is updated.

rasika-chivate:
> ⚠️ the bigger problem is: As for now, our Doc will have instructions on how to update Binary, but we don't have "update pmm client by packages" (our primary method).
> 
> I see we have https://www.percona.com/doc/percona-monitoring-and-management/2.x/setting-up/client/index.html#binary-package and the same for binary, so we can reference them if needed. But we must have some short instructions for the update by packages near the binaries update

@rnovikovP Absolutely we should update the doc, but I think we should handle this as part of a separate JIRA ticket. Please confirm and then I will create a JIRA ticket for this change.

denisok (approved):

rasika-chivate (approved):

PR #664, opened 2021-12-16
stop hinting at port 80 use!

shoffman-percona:
Removed reference to HTTP use as the way that was worded originally made it sound like if port 80 was enabled you should use that instead...


rnovikovP (approved):

rasika-chivate (commented):
Rephrasing a bit:

Visit `https://localhost:443` to see the PMM user interface in a web browser. (If you are accessing the docker host remotely, replace `localhost` with the IP or server name of the host.)

PR #676, opened 2022-01-04
PMM-8496 change quick install

nikita-b:
I slightly changed quick install:
1. We still use GitHub for storing bash script and I think it's the right way:
- We don't need to maintain script in our infrastructure
- We'll store the script in git and can see history of changes

2. I removed SHA check but I think we can add it on our side (one time in 5 minutes for example). I'll ask Percona IT.

3. I changed the wording a little.


rasika-chivate (changes requested):
This script:

- Installs Docker if it is not installed.
- If a PMM Server Docker container is running, it is stopped and backed up.
- Pulls and runs the latest PMM Server Docker image.

atymchuk (approved):

rasika-chivate (approved):

PR #685, opened 2022-01-18
PMM-9336 Add AWS access key ID and secret access key

tiagomotasantos:
Adds AWS access key ID and secret access key to Kubernetes register modal so users that are using EKS and this values are not present in kubeconfig can fill them. If those users don't fill these fields an error is thrown since the cluster register fails.

https://jira.percona.com/browse/PMM-9336


rasika-chivate (commented):
Made some minor changes.

rasika-chivate (approved):

PR #697, opened 2022-02-04
Remove docker compose from pmm-client

rnovikovP:
@rasika-chivate  pls 👀  as we might have docker-compose in some other places too. 
Until we are ready to support it - it can be in the official documentation. 
The current examples are outdated and limited, so they are just confusing. 

pls clean up this PR, I'm sure more removals are needed


rasika-chivate:
@rnovikovP I have removed references of docker-compose from the PMM Client.

denisok (approved):
I don't mind to review it. One reason for it to stay - it is probably good example for the ppl that use docker-compose for all other services (DBs) and would like to run our client in the same environment.
So this is a good example how to do that. Otherwise ppl would need to dig into the details what env they should use and etc

rnovikovP:
> this is a good example of how to do that

We can move this to Forum or make a blog post, which is ok with me. 
The only problem is that this is not updated and in the official documentation.

BupycHuk (approved):

rasika-chivate:
> > this is a good example of how to do that
> 
> We can move this to Forum or make a blog post, which is ok with me. The only problem is that this is not updated and in the official documentation.

@rnovikovP Should I merge it and publish it...

rnovikovP:
@rasika-chivate, let me think about it. We got some ideas about centralized place for such articles... I'll handle it

edithturn:
Hello @rnovikovP ! I would like to help with docker-compose, I was looking for a way to start everything pmm-client, pmm-server, and pmm-server-server with a single command: `docker-compose` up, to make fast tests. If you need help to test or update the documentation please, let me know.
I was testing a single Docker compose file joining the docker installation of pmm-client and pmm-server, but this is not working when I connect a database. I tried to connect them through the container port without success.

` pmm-server:
    container_name: pmm
    image: percona/pmm-server:2
    platform: linux/amd64
    ports:
      - 32433:80
    volumes:
      - pmm-data:/srv
    healthcheck:
      test: curl localhost/server/login | grep Unauthorized
      interval: 1s
      retries: 100
      start_period: 1s
      timeout: 10s

  pmm-client:
    image: percona/pmm-client:2
    hostname: pmm-client-myhost
    container_name: pmm-client
    restart: always
    ports:
      - "42000:42000"
      - "42001:42001"
    logging:
      driver: json-file
      options:
        max-size: "10m"
        max-file: "5"
    volumes:
      - ./pmm-agent.yaml:/etc/pmm-agent.yaml
      - pmm-client-data:/srv
    environment:
      - PMM_AGENT_CONFIG_FILE=/etc/pmm-agent.yaml
      - PMM_AGENT_SERVER_USERNAME=admin
      - PMM_AGENT_SERVER_PASSWORD=admin
      - PMM_AGENT_SERVER_ADDRESS=172.21.0.2:443
      - PMM_AGENT_SERVER_INSECURE_TLS=true
    entrypoint: pmm-agent setup

volumes:
  pmm-data:
  pmm-client-data:`

Let me know if you need any more information about the compose and if I can help you with it 😃

rnovikovP:
@edithturn, thanks for the comment (I missed it initially ((  )  
The goal now - is to remove the "deployment method we never intend to be added" as this is not something the Engineering team was ready to support. 
We can keep this in Blog posts but not in official documentation, as we have no process to support this method up to date. 
Example: ports on pmm-client are not something you need to expose in current.

PR #710, opened 2022-02-16
PMM-9306 use latest VM version for PMM

rasika-chivate:
Add a note to use the latest version for Victoria Metrics for your PMM instance.


rnovikovP:
PMM users can't control the version of VM they use. This is in PMM control, so we should not recommend updating versions separately. 


We used some "cool version" of one of our components that should be reflected as value to the end-user "faster, more stable, etc.". We also should have some Performance tests to confirm  (@puneet0191, ptal . do we have this? )

If we can't explain and show the real value to the end-user, there is not much we can advertise for them.

rnovikovP (changes requested):
see the comment

rasika-chivate:
This will be a part of release notes.

PR #721, opened 2022-03-04
Add advisors to the docs

atymchuk:


Tusamarco (approved):
As for chat with Roma I ack this PR but keep in mind we have same set of files in PORTAL-DOC

PR #747, opened 2022-04-01
Update conf-mongodb.md

igroene:
Let's make this consistent with https://www.percona.com/doc/percona-monitoring-and-management/2.x/setting-up/client/mongodb.html


atymchuk (approved):

igroene:
actually the new dbstas and colstats functionality of mongodb_exporter require that we give the perms on all collections, so the page as it was is correct now.

PR #757, opened 2022-04-11
PMM-9331 mysql commands for user creation

rasika-chivate:
Modified the CREATE DATABASE command for MySQL 5.7


puneet0191 (approved):

rnovikovP:
LGTM, but let's put MySQL 8 section first, as it's the more recent version, and the command is more complex. 
So, if a user with mysql8 executes the First commands (mysql5.7 ), they will get OK and have problems later. 
If we put  mysql8 first, then 5.7 users executing this command will get an error, forcing them to read carefully and execute the next section of commands.

denisok (approved):

rnovikovP (approved):
only  https://github.com/percona/pmm-doc/pull/757#issuecomment-1094887084

PR #799, opened 2022-05-19
PMM-9590 credentials source flag

rasika-chivate:
Added credentials-source flag which can be used instead of username and password flags.


denisok (approved):

rnovikovP (changes requested):
This now looks like Auth command for `pmm-admin` in general. 

But in reality, this is part of `add` command.
I think this needs to be moved to   Database commands  section 

image

beatahandzelova (approved):

beatahandzelova (approved):

rnovikovP (approved):

PR #805, opened 2022-05-25
PMM-9354 external clickhouse database instance

rasika-chivate:
Ability to add external clickhouse database instance for PMM server.


rasika-chivate:
Fix version is 2.29.0 so have to change the base to 2.29.0

PR #815, opened 2022-07-01
PMM-9632 pg_stat_monitor settings reflect.

JiriCtvrtka:


denisok (approved):

docs/setting-up/client/postgresql.md
```diff
@@ -190,6 +190,13 @@ You can now [add the service](#add-service).
 
     You can get a list of other available settings with `SELECT * FROM pg_stat_monitor_settings;`.
 
+    Another useful settings item is
```
rasika-chivate, line 193:
```suggestion
    Another useful settings parameter is:
```

docs/setting-up/client/postgresql.md
```diff
@@ -190,6 +190,13 @@ You can now [add the service](#add-service).
...
+    ```ini
+    pg_stat_monitor.pgsm_normalized_query
+    ```
+
+    If value is set to 1 then query acutal values are replaced by placeholders. If value is 0 then you will be able to see examples in QAN.
```
rasika-chivate, line 198:
```suggestion
     If the value is set to 1, then query actual values are replaced by placeholders. If the value is 0, you can see the examples in QAN. 
```

rasika-chivate (changes requested):

docs/setting-up/client/postgresql.md
```diff
@@ -190,6 +190,13 @@ You can now [add the service](#add-service).
 
     You can get a list of other available settings with `SELECT * FROM pg_stat_monitor_settings;`.
 
+    Another useful settings parameter is:
```
rasika-chivate, line 193:
```suggestion
    Another important parameter is:
```

docs/setting-up/client/postgresql.md
```diff
@@ -190,6 +190,13 @@ You can now [add the service](#add-service).
...
+    ```ini
+    pg_stat_monitor.pgsm_normalized_query
+    ```
+
+     If the value is set to 1, then query actual values are replaced by placeholders. If the value is 0, you can see the examples in QAN.
```
rasika-chivate, line 198:
```suggestion
    If the value is set to 1, the actual query values are replaced by placeholders. If the value is 0, the examples are given in QAN. 
```

rasika-chivate: Can we give a link to QAN where the examples are there?

rasika-chivate (changes requested):

docs/setting-up/client/postgresql.md
JiriCtvrtka: Examples are in detail of specific row, so direct link is always different depends on opened row. If you want we can describe it somehow. Its when you open row details then there is tab called examples.

JiriCtvrtka: @rasika-chivate Take a look on message above. Is it OK?

rasika-chivate: I think we should describe it.

jctvrtka: Examples can be found in QAN details tab example.

Is this OK? Or better ideas how to describe it?

JiriCtvrtka: Added, resolved.

PR #816, opened 2022-07-01
Update check development pages to note that custom check development …

mbenshoof:
…is currently debugging/testing only.

Documentation should note that local checks are for testing/debugging only and aren't intended for production use.  In future versions, when local checks are fully supported, the documentation should be updated to remove this limitation.


denisok:
@Hlmn-Rmr please review

PR #846, opened 2022-08-10
PMM-10378 simple k8s registration

rasika-chivate:
Simple registration for Kubernetes cluster


denisok (changes requested):
there are couple of screenshots with "unsupported" operator. Those should be replaced with normal mode "green" operators.

denisok (approved):

solovevayaroslavna (approved):

dygksy (approved):

docs/using/dbaas.md
```diff
@@ -35,7 +35,10 @@ From the left menu, select <i class="uil uil-database"></i> *DBaaS*.
 
 1. Click *Register new Kubernetes Cluster*.
 
-2. Enter values for the *Kubernetes Cluster Name* and *Kubeconfig file* in the corresponding fields.
+2. Copy the value of *Kubeconfig file* and click *Paste from clipboard* to copy the content of the kubeconfig file in the corresponding field. The value of *Kubernetes Cluster Name* gets auto-populated from the contents of the *kubeconfig file*.
```
beatahandzelova, line 38: Paste from clipboard feature doesn't work in some cases. Maybe we should mention this in docs too?

https://developer.mozilla.org/en-US/docs/Web/API/Clipboard
https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#browser_compatibility

solovevayaroslavna (changes requested):
@rasika-chivate as Beata mentioned will better to mention that copytoclipboard API doesn't work in some cases.

docs/using/dbaas.md
rasika-chivate: @beatahandzelova what are the cases where this does not work.

beatahandzelova: @rasika-chivate it's all in the links above:
This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#browser_compatibility).

rasika-chivate: @beatahandzelova @solovevayaroslavna which browsers are supported. We should also mention that...

beatahandzelova: I'd just use the link [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#browser_compatibility) because the supporting browsers/versions might change in future.

beatahandzelova (approved):

rnovikovP (approved):

solovevayaroslavna (approved):

PR #873, opened 2022-09-21
Percona alerting

catalinaadam:


docs/faq.md
```diff
@@ -99,18 +99,14 @@ The default values (in seconds):
 
 ## How do I set up Alerting?
 
-When a monitored service metric reaches a defined threshold, PMM Server can trigger alerts for it either using the Grafana Alerting feature or by using an external alert manager.
+When a monitored service metric reaches a defined threshold, PMM Server can trigger alerts for it using embedded Grafana Alerting functionality.
 
-With these methods you must configure alerting rules that define conditions under which an alert should be triggered, and the channel used to send the alert (e.g. email).
+For this, you must configure alerting rules that define conditions under which an alert should be triggered, and the communication points used to send the alert (e.g. email).
```
artemgavrilov, line 104:
```suggestion
For this, you must configure alerting rules that define conditions under which an alert should be triggered, and the contact points used to send the alert (e.g. email).
```

artemgavrilov: That's how it's named in Grafana

docs/using/alerting.md
```diff
@@ -136,46 +124,62 @@ For example, to validate that a MongoDB instance is down, shut down a member of
...
+### Add an alert rule
+After provisioning the resources required for creating Percona templated alerts, you are now ready to create your alert:
 
-If you’re migrating from PMM 2.25 and earlier, make sure to manually recreate any custom alert rules and rule templates that you want to transfer to PMM 2.26.0 or later.
+1. Go to **Alerting > Alert Rules**, and click **New alert rule**.
+2. On the **Create alert rule** page, select the **Percona templated alert** option. If you want to learn about creating Grafana alerts instead, check our [Grafana's documentation](https://grafana.com/docs/grafana/latest/alerting/).
+3. In the **Template details** section, choose the template on which you want to base the new alert rule. This automatically populates the **Name**, **Duration**, and **Severity** fields with information from the template. You can change these values if you want to override the default specifications in the template.
+4. In the **Filters** field, specify if you want the alert rule to apply only to specific services or nodes. For example: `service_name'`, Operator:`=(EQUAL)`, VALUE: `ps5.7`.
```
artemgavrilov, line 133:
```suggestion
4. In the **Filters** field, specify if you want the alert rule to apply only to specific services or nodes. For example: `service_name'`, Operator:`MATCH`, VALUE: `ps5.7`.
```

artemgavrilov: That part was changed, now we have two filters types: `MATCH` and `MISMATCH`, both works as a RegEx.

artemgavrilov (approved):

docs/using/alerting.md
```diff
@@ -122,60 +129,80 @@ templates:
...
+2. Click **Add** and upload a local .yaml file from your computer.
+
+#### Migrate alert rules
+Alert rules created with Integrated Alerting in PMM 2.30 and earlier are not automatically migrated to Percona Alerting. 
+
+After upgrading to PMM 2.31, make sure to manually migrate any alert rules that you want to transfer to PMM 2.31 using the [Integrated Alerting Migration Script](https://github.com/percona/pmm/blob/main/ia_migration.py).
+
+The default command for migrating rules is:
+```yaml 
+*python migration.py -u admin -p admin*
```
artemgavrilov, line 190:
```suggestion
*python ia_migration.py -u admin -p admin*
```

artemgavrilov: I guess we can add something like 

```
To see all available options check scrip help `ia_migration.py -h`
```

docs/using/alerting.md
```diff
@@ -122,60 +129,80 @@ templates:
...
+```
+##### Script prerequisites
+- Python version 3.x, which you can download from [Python Downloads centre](https://www.python.org/downloads/).
+  
+-  [Requests  library](https://requests.readthedocs.io/en/latest/user/install/#install), which you can install with the following command: ```pip install requests```.
```
artemgavrilov, line 195: @gen1us2k please check if it's enough

PR #886, opened 2022-10-14
Added ref to enabling pgsm_enable_query_plan.

guriandoro:
With pg_stat_monitor.pgsm_enable_query_plan we can enable a very important feature of pg_stat_monitor, that is saving explain plans for queries. They can be viewed in the QAN dashboard, in the "Plan" tab (which is something we don't have with pg_stat_statements).

Additionally, fixed broken link to github page referring to available parameters to tune, which now points to our online docs.


guriandoro:
Created a new one that is a bit less verbose.

PR #920, opened 2022-11-10
Fix supported MySQL backup setups

sleto-it:
Follow-up of PR #917 & PR #916.

The net effect of these two PRs about MySQL supported setups was that we have two sections for MySQL.

This PR fixes the situation by:

- removing one duplicate section
- making sure S3-compatible is still mentioned

Thanks


sleto-it:
**Before (two sections):**

before-two-sections

**After the fix:**

after-the-fixes

catalinaadam (approved):

PR #923, opened 2022-11-18
Update mongodb_limitations.md

mmnosek:
I find it really difficult to read that massive table with what is supported and what is not. I suggest we remove two columns and do 4 shorter tables instead. That way it's much easier to navigate and find what you need.


rnovikovP (approved):
Thanks, @mmnosek ! 
indeed this is a better presentation if you  say so ) 

I'm thinking about another form, but let's simplify what we know is better for now; we can always improve in the future.

catalinaadam (approved):

rasika-chivate (approved):

PR #940, opened 2022-12-12
Update pmm_components_and_versions.md

rasika-chivate:


BupycHuk (approved):

atymchuk (approved):
@rasika-chivate Just found something that doesn't read well:
`v0.14.0 and Percona changes`

Maybe mark it with a star and than provide a legend right below?
* - Original upstream version along with some changes authored by Percona

Would it work?

PR #960, opened 2023-01-24
Update dbaas.md

mmnosek:
Clarifying what clusters are supported currently in DBaaS, as many people tried e.g. with GKE and it fails. Worth mentioning what's expected to work for clarity.


rnovikovP:
We also need @PeterSzcz here as PM for this .. but I can't add him as a reviewer(

gen1us2k (approved):

PR #962, opened 2023-01-25
PMM-11445 edit page missing

rasika-chivate:
Edit page missing on PMM docs

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -6,14 +6,7 @@ site_url: https://docs.percona.com/percona-monitoring-and-management/
 plugins:
   section-index: {}
 
-features:
-  - search.highlight
-  - navigation.top
-  - navigation.tracking 
-  - content.tabs.link
-  - content.action.edit
-  - content.action.view
-  - content.code.copy 
+ 
 
 #Google Analytics configuration
 extra:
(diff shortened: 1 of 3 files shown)


LenzGr:
@rasika-chivate please consider adding some background info to the git commit message about this change, and why it is necessary. I added the following to mine:
```

Material Theme for MkDocs version 9.x changed the default behavior of the
"Edit this page" functionality. Now both editing and viewing the source code
must be explicitly enabled:

https://squidfunk.github.io/mkdocs-material/upgrade/#upgrading-from-8x-to-9x
```

atymchuk:
@rasika-chivate @LenzGr Why don't we pin the MaterialUI version so we 1) upgrade when necessary and 2) don't face surprises like this.

atymchuk (approved):

mkdocs.yml
```diff
@@ -6,6 +6,15 @@ site_url: https://docs.percona.com/percona-monitoring-and-management/
 plugins:
   section-index: {}
 
+features:
```
atymchuk, line 9: I believe this changes belongs in mkdocs-base.yml, since it is extended by multiple yml files.

atymchuk (changes requested):

atymchuk (approved):

LenzGr:
@rasika-chivate seems like the edit icons re-appeared on the PMM docs, even though this PR hasn't been merged. Could you please take a look?

rasika-chivate:
Modified the docs with the PR: https://github.com/percona/pmm-doc/pull/972

PR #1016, opened 2023-03-11
Fix broken link to public address

atymchuk:
The link to public address section was broken

Screenshot 2023-03-11 at 14 51 12


atymchuk:
@rasika-chivate Why do we have all pages rendered four times? Does this add any more value compared to just one? Which one is the right one?

rasika-chivate (approved):

PR #1052, opened 2023-04-12
correcting misinformation

shoffman-percona:
That line hinted that we did some sort of backup and we don't...if there's a container named "pmm-server" we simply stop it, and rename it so we can create another pmm-server container of the latest version.  

I don't want people to think we take any kind of recoverable backup if you want to go back to the old version...that won't work.


rasika-chivate (approved):

docs/setting-up/server/easy-install.md
```diff
@@ -17,7 +17,7 @@ wget -O - https://www.percona.com/get/pmm | /bin/bash
 This script:
 
 - Installs Docker if it is not already installed on your system.
-- Stops and backs up any PMM Server Docker containers that are currently running.
+- Stops and renames any PMM Server Docker container found to be running (not a recoverable backup being taken)
```
catalinaadam, line 20:
```suggestion
- Stops and renames any PMM Server Docker container currently running from `pmm-server` to `pmm-server-backup-{timestamp}`. This new pmm-server container based on the latest version is not a recoverable backup.
```

catalinaadam (commented):
Not sure my suggestion is correct but we could probably be more specific here. @idoqo, could you please check my suggestion and advise?

docs/setting-up/server/easy-install.md
idoqo: instead of 

"This new pmm-server container based on the latest version is not a recoverable backup.",

Can it be "This old pmm-server container is not a recoverable backup."?

idoqo (approved):

catalinaadam (approved):

catalinaadam (approved):

catalinaadam (approved):

PR #1060, opened 2023-04-20
Update FAQ with questions related to personal data

rasika-chivate:
Add questions related to personal data collection as per request:

https://docs.google.com/document/d/1cImrLRu8AATxhTOHiCRQfheBcBub7ho4p0Zh21bdCuM/edit?pli=1


docs/faq.md
```diff
@@ -184,4 +184,10 @@ If you are using OVF/AMI, you can change the default password through SSH by usi
 
 ```sh
 change-admin-password <new_password>
-```
\ No newline at end of file
+```
+
+
```
denisok, line 189: too many new lines?

rasika-chivate: Removed

docs/personal_data_handling.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+The following questions are being answered related to personal and confidential data handling in PMM:
+
+1. Which type of data is transmitted?
+
+      |**Data collection source**                                       | **Data collected** |
+      | --------------------------------------------------------------- | ------------------------------------------------------
+      | DB host to PMM                                                  | Database performance metrics <br/> SQL query examples for query analytics (optional).
+      | PMM to DB Host                                                  | DSN and credentials for database access. A separate DB user is used (read-only) to access the database.
```
shoffman-percona, line 11: Is this true?  I seem to recall you need some amount of privs to flush/rotate either perf schema or slowlog... @rnovikovP am I off here?  I think "limited access account" might be a better fit if so.

docs/personal_data_handling.md
```diff
@@ -0,0 +1,56 @@
...
+      | DB Host to Cloud S3                                             | Database backup - optional if PMM Administrator configures it.
+      | PMM Server to Percona Cloud                                     | Telemetry data is collected. </br/> PMM Server collects varying amounts of data from version to version. <br/> No personal or confidential information is collected.
+
+
+2. Where is the data transmitted?
```
shoffman-percona, line 16: Not sure what data we're talking about here...telemetry also transmits to percona cloud...

docs/personal_data_handling.md
```diff
@@ -0,0 +1,56 @@
...
+
+3. What is the purpose and nature of data processing?
+
+    As per our [Privacy Policy](https://www.percona.com/privacy-policy), the data collection purposes are to provide the services and product enhancements.
+
+    Although, PMM does not collect nor transfer personal data explicitly, in case query analytics is enabled and query examples collection is not disabled, we gather SQL query examples with real data and personal data may appear there if it is stored in DB.  All QAN data always remains within the PMM server, and is never transmitted anywhere else.
+
+4. What is the frequency and volume of processed data?
+
+    Data is gathered each minute. In case Query Analytics is enabled and SQL query examples are gathered, we don't use any special processing for personal or confidential data. PMM server has no clue about the meaning of the data inside the SQL query.
```
shoffman-percona, line 30:
```suggestion
    While this is user configurable, by default data is gathered every 5, 10 or 60 minutes. In case Query Analytics is enabled and SQL query examples are gathered, we don't use any special processing for personal or confidential data. PMM server has no clue about the meaning of the data inside the SQL query. 
```

docs/personal_data_handling.md
```diff
@@ -0,0 +1,56 @@
...
+
+    As per our [Privacy Policy](https://www.percona.com/privacy-policy), the data collection purposes are to provide the services and product enhancements.
+
+    Although, PMM does not collect nor transfer personal data explicitly, in case query analytics is enabled and query examples collection is not disabled, we gather SQL query examples with real data and personal data may appear there if it is stored in DB.  All QAN data always remains within the PMM server, and is never transmitted anywhere else.
+
+4. What is the frequency and volume of processed data?
+
+    Data is gathered each minute. In case Query Analytics is enabled and SQL query examples are gathered, we don't use any special processing for personal or confidential data. PMM server has no clue about the meaning of the data inside the SQL query. 
+
+    So it is processed as usual, which is to store inside the PMM server and present on the PMM UI by request.
```
shoffman-percona, line 32: should this be part of the previous paragraph?

docs/personal_data_handling.md
```diff
@@ -0,0 +1,56 @@
...
+
+    Other than email addresses for Grafana users, PMM does not directly ask or collect any other personal data. For more information about the telemetry data that is collected, please refer to the [Percona Privacy Policy](http://www.percona.com/privacy-policy/). 
+
+5. What applications or third parties can access the data created and processed by the cloud service?
+
+    Third parties or other applications are not able to access the data gathered by the PMM server.
+
+    Although PMM does not collect nor transfer personal data explicitly, in case query analytics is enabled and query examples collection is not disabled, PMM gathers SQL query examples with real data and personal data may appear there if it is stored in DB.  All QAN data always remains within the PMM server, and is never transmitted anywhere else
```
shoffman-percona, line 40: this doesn't make sense here...and we already covered it in #3 above.

docs/personal_data_handling.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+6. Is Personal Data processed for other applications or parties, and should the data that is processed in the cloud service be available to other applications or 3rd parties?
+
+    PMM server doesn't pass any gathered, personal or confidential data to any third party or other applications nor to Percona Cloud.
+
+7. How safe is the encryption? It's a must to encrypt all connections to and from the cloud including the data in the cloud storage.
```
shoffman-percona, line 47:
```suggestion
7. How safe is the encryption? 

It's a must to encrypt all connections to and from the cloud including the data in the cloud storage and PMM does so by default. 
```

shoffman-percona (changes requested):
Few questions and some suggestions for clarity.

docs/personal_data_handling.md
maxkondr: here DSN and credentials are referred to "limited access account" (pmm), but anyway they are passed to pmm-agent so that it is able to connect to DB.

docs/personal_data_handling.md
```diff
@@ -0,0 +1,56 @@
...
+
+The following questions are being answered related to personal and confidential data handling in PMM:
+
+1. Which type of data is transmitted?
+
+      |**Data collection source**                                       | **Data collected** |
+      | --------------------------------------------------------------- | ------------------------------------------------------
+      | DB host to PMM                                                  | Database performance metrics <br/> SQL query examples for query analytics (optional).
+      | PMM to DB Host                                                  | DSN and credentials for database access. A separate DB user is used (read-only) to access the database.
+      | DB Host to Cloud S3                                             | Database backup - optional if PMM Administrator configures it.
```
maxkondr, line 12: need to mention that this is not Percona Cloud, this is S3 storage that PMM Administrator wants to use as DB backup storage. It can be any S3 in public clouds (AWS,GCP, etc) or internal S3-compatible storage

docs/personal_data_handling.md
BupycHuk: @shoffman-percona, it's not, we don't create a separate DB user. So we use credentials passed by the user.

docs/personal_data_handling.md
BupycHuk, line 16: seems like 1st paragraph answers to this question

docs/personal_data_handling.md
BupycHuk: if we are talking about QAN, it's not configurable.

docs/personal_data_handling.md
```diff
@@ -0,0 +1,56 @@
...
+
+6. Is Personal Data processed for other applications or parties, and should the data that is processed in the cloud service be available to other applications or 3rd parties?
+
+    PMM server doesn't pass any gathered, personal or confidential data to any third party or other applications nor to Percona Cloud.
+
+7. How safe is the encryption? It's a must to encrypt all connections to and from the cloud including the data in the cloud storage.
+
+    We use TLS for connections between:
+
+    - Database host to PMM Server
```
BupycHuk, line 51: We use it only if it was configured by PMM Administrator

docs/personal_data_handling.md
shoffman-percona: Yes but the idea is that PMM collects and stores this data (plain text) in it's master database (pg) so we need to disclose that to the users...my question was more about the privs required for "best practice" configuration...is it truly a "read-only" user we ask them to create or does that user have some ability to modify a remote system?

BupycHuk: it's not read-only. as you can see in following links we request more permissions then just read.
https://docs.percona.com/percona-monitoring-and-management/setting-up/client/mongodb.html#create-roles-with-privileges-for-backups-and-qan
https://docs.percona.com/percona-monitoring-and-management/setting-up/client/postgresql.html#create-a-database-account-for-pmm
https://docs.percona.com/percona-monitoring-and-management/setting-up/client/mysql.html#create-a-database-account-for-pmm
I would recommend not mentioning it here at all. Because it seems like we create a separate DB user.

shoffman-percona: perfect...we'll remove the read-only part but we do need to disclose that we are capturing and storing this user-created db users credentials in PMM along with DSN.

shoffman-percona (approved):
Looks good!

PR #1099, opened 2023-06-21
PMM_11821_Telemetry

rasika-chivate:
https://jira.percona.com/browse/PMM-11821


docs/how-to/configure.md
```diff
@@ -103,9 +93,36 @@ Grafana’s [anonymous usage statistics](https://grafana.com/docs/grafana/latest
...
+    Change interval for telemetry iteration
+
+
+For information on the various config parameters for telemetry, see the [config file](https://github.com/percona/pmm/blob/main/managed/services/telemetry/config.default.yml).
+
+
 !!! note alert alert-primary ""
 
-    Telemetry is sent straight away; the 24 hour grace period is not honored.
+    PMM retains the telemetry data by default for 30 days. However, this can be modified by an admin from the settings page.
```
rnovikovP, line 124: I'm not able to get the context here. if this is to PERCONA_TEST_*  this  need to be clarified, but I would avoid PERCONA_TEST_ in official doc at all

docs/how-to/configure.md
```diff
@@ -103,9 +93,36 @@ Grafana’s [anonymous usage statistics](https://grafana.com/docs/grafana/latest
...
 
+The following are some of the commonly used environment varibles for telemetry:
+
+```sh
+DISABLE_TELEMETRY=true
+```
+      Disable the telemetry feature
+
+```sh
+PERCONA_TEST_TELEMETRY_DISABLE_SEND=true
```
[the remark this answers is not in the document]
rasika-chivate: ok. @maxkondr I think we should remove it. Please confirm.

docs/how-to/configure.md
maxkondr: Wait, 30 days relate to metrics in general. 
We don't have such constrains for Telemetry data at all.

rnovikovP: This is why I think this line should stay as is. No changes are needed here.

rnovikovP (changes requested):

docs/how-to/configure.md
```diff
@@ -73,21 +73,15 @@ Values for the *Custom* preset can be entered as values, or changed with the arr
...
-- PMM Server Check Updates feature disabled
-- Detailed information about the version of monitored MySQL services
-- Monitored MongoDB services version
-- Monitored PostgreSQL services version
-- Total Grafana users
-- Monitored nodes count
-- Monitored services count
-- Agents version
-- Node type
+PMM collects data through exporters and Query Analytics, which is then utilized by the Telemetry subsystem through aggregated metrics.
```
rnovikovP, line 76:
```suggestion
PMM Telemetry is based on data collected by various PMM components and stored inside PMM server 
```
no need to mention components as they might change. and actually, we use not only exporters and qan but some grafana and UI data.

docs/how-to/configure.md
```diff
@@ -73,21 +73,15 @@ Values for the *Custom* preset can be entered as values, or changed with the arr
...
-- Monitored PostgreSQL services version
-- Total Grafana users
-- Monitored nodes count
-- Monitored services count
-- Agents version
-- Node type
+PMM collects data through exporters and Query Analytics, which is then utilized by the Telemetry subsystem through aggregated metrics. 
+
+-!!! note alert alert-primary ""
+    Telemetry is sent straight away. The 24-hour grace period is not applicable.
```
rnovikovP, line 79:
```suggestion
    Telemetry is not sent straight away ante the PMM is installed. PMM has a 24-hour grace period before the first telemetry report to allow users to disable telemetry over UI.
```

rnovikovP (changes requested):

rnovikovP (approved):

PR #1127, opened 2023-07-28
Improvements from HD-20808

catalinaadam:


mkdocs.yml
```diff
@@ -13,3 +13,9 @@ extra:
   analytics:
     provider: google
     property: UA-343802-3
+
+
+plugins:
+  section-index: {}
```
alina-derkach-oaza, line 19: Be careful with this plugin. It can break the side TOC.

PR #1129, opened 2023-07-31
PMM-6279 Query metadata not visible in new UI

rasika-chivate:
Added metadata table for a query in QAN. For details see,

https://jira.percona.com/browse/PMM-6279


rnovikovP (approved):
LGTM,  maybe we need a better explanation of the metadata, but we can add it later based on feedback

JiriCtvrtka (approved):

PR #1150, opened 2023-09-27
PMM-12500 k8s dashboard

rasika-chivate:
For details see,

https://jira.percona.com/browse/PMM-12500


rnovikovP (changes requested):
1) this should be a 2 separate pages, as there 2 dashboards inside 
2) there is the same image for both dashboards, pls check it's should be 2 different images

rnovikovP (approved):

PR #1155, opened 2023-10-05
PMM-7 upgrade dependencies

ademidoff:
This was to see if we need to keep the Markdown version pinned for longer. It turns out the problem with incompatible dependecies we faced a while ago was solved.

The preview being compiled just fine, let's get rid of the pin :)


alina-derkach-oaza (approved):
LGTM

rasika-chivate (approved):

PR #1165, opened 2023-10-16
PMM-12544 : DbaaS warning

catalinaadam:


matejkubinec:
Looks good to me.

Are the links to migration guide/everest final?

matejkubinec (approved):

rnovikovP (changes requested):
Everest - http://per.co.na/pmm-to-everest 
guide - http://per.co.na/pmm-to-everest-guide 

We must use the short links here, not the destinations as the links are not final yet.

matejkubinec (approved):

PR #1181, opened 2023-12-06
Revert changes from 2.41.0

rasika-chivate:
Revert PMM-8655 from PMM 2.41.0.


rasika-chivate:
Need to change the base so clsoing the PR and creating a new PR.

PR #1229, opened 2024-04-05
Update mongodb.md

igroene:
Clarify what to do when enabling profiler dynamically and other fixes

### DIFF
--- a/docs/setting-up/client/mongodb.md
+++ b/docs/setting-up/client/mongodb.md
@@ -194,7 +194,7 @@ For example, to enable the profiler in the `testdb`, run this:
 
 ## Add service
 
-When you have configured your database server, you can add a MongoDB service with the user interface or on the command line.
+"After configuring your database server, you can add a MongoDB service either through the user interface or via the command line.
 
 !!! caution alert alert-warning "Important"
     To monitor MongoDB sharded clusters, PMM requires access to all cluster components. Make sure to add all config servers, shards, and at least 1-2 mongos routers (adding all mongos routers may cause too much overhead). Otherwise, PMM will not be able to correctly collect metrics and populate dashboards.
--- a/docs/setting-up/client/mongodb.md
+++ b/docs/setting-up/client/mongodb.md
@@ -141,7 +141,7 @@ You can set profiling:
 - until the next database instance restart, by running a command in a `mongo` session.
 
 !!! note alert alert-primary ""
-    Profiling is turned off by default as it can adversely affect the performance of the database server.
+    Profiling is turned off by default as it can adversely affect the performance of the database server on certain circumstances (e.g. busy servers profiling all queries).
 
 ### Set profiling in the configuration file
 
@@ -190,14 +190,14 @@ For example, to enable the profiler in the `testdb`, run this:
 ```
 
 !!! note alert alert-primary ""
-    If you have already [added a service](#add-service), you should remove it and re-add it after changing the profiling level.
+    If you have already [added the MongoDB service to PMM](#add-service), you should restart the PMM agent service after changing the profiling level.
 
 ## Add service
 
 When you have configured your database server, you can add a MongoDB service with the user interface or on the command line.
 
 !!! caution alert alert-warning "Important"
-    To monitor MongoDB sharded clusters, PMM requires access to all cluster components. Make sure to add all config servers, shards, and mongos. Otherwise, PMM will not be able to correctly collect metrics and populate dashboards.
+    To monitor MongoDB sharded clusters, PMM requires access to all cluster components. Make sure to add all config servers, shards, and at least 1-2 mongos routers (adding all mongos routers may cause too much overhead). Otherwise, PMM will not be able to correctly collect metrics and populate dashboards.
 
 ### With the user interface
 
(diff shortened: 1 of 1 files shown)


JiriCtvrtka (approved):

docs/setting-up/client/mongodb.md
```diff
@@ -190,14 +190,14 @@ For example, to enable the profiler in the `testdb`, run this:
 ```
 
 !!! note alert alert-primary ""
-    If you have already [added a service](#add-service), you should remove it and re-add it after changing the profiling level.
+    If you have already [added the MongoDB service to PMM](#add-service), you should restart the PMM agent service after changing the profiling level.
```
BupycHuk, line 193: 👍🏼

docs/setting-up/client/mongodb.md
```diff
@@ -217,7 +217,8 @@ Use `pmm-admin` to add the database server as a service using one of these examp
...
 
 !!! hint alert alert-success "Tips"
-    - When adding nodes of a sharded cluster, add each node separately using the `--cluster mycluster` option for the [MongoDB Cluster Summary](../../details/dashboards/dashboard-mongodb-cluster-summary.md) dashboard to populate correctly. Also use the `--replication-set` option to specify a replication set. Example: `--replication-set config` for your config servers; `--replication-set rs1` for your servers in the first replica set, `--replication-set rs2` for your servers in the second replica set, and so on.
+    - When adding nodes of a sharded cluster, add each node separately using the `--cluster mycluster` option for the [MongoDB Cluster Summary](../../details/dashboards/dashboard-mongodb-cluster-summary.md) dashboard to populate correctly. 
+    - You can also use the `--replication-set` option to specify a replication set, altough they are automatically detected. Example: `--replication-set config` for your config servers; `--replication-set rs1` for your servers in the first replica set, `--replication-set rs2` for your servers in the second replica set, and so on.
```
BupycHuk, line 221: > altough they are automatically detected

Are they? Most of dashboards keep working, because we just use another label. But for backups as far as I remember it's critical to pass that flag.

igroene: well, we can derive the replication set name from metrics, so it should be auto detected. Otherwise the user could supply the wrong replication-set which would cause all sorts of problems.

docs/setting-up/client/mongodb.md
catalinaadam, line 193:
```suggestion
    If you have already [added the MongoDB service to PMM](#add-service), make sure to restart the PMM agent service after adjusting the profiling level.
```

docs/setting-up/client/mongodb.md
```diff
@@ -190,14 +190,14 @@ For example, to enable the profiler in the `testdb`, run this:
 ```
 
 !!! note alert alert-primary ""
-    If you have already [added a service](#add-service), you should remove it and re-add it after changing the profiling level.
+    If you have already [added the MongoDB service to PMM](#add-service), you should restart the PMM agent service after changing the profiling level.
 
 ## Add service
 
 When you have configured your database server, you can add a MongoDB service with the user interface or on the command line.
```
catalinaadam, line 197:
```suggestion
"After configuring your database server, you can add a MongoDB service either through the user interface or via the command line.
```

docs/setting-up/client/mongodb.md
```diff
@@ -217,7 +217,8 @@ Use `pmm-admin` to add the database server as a service using one of these examp
 When successful, PMM Client will print `MongoDB Service added` with the service's ID and name. Use the `--environment` and `-custom-labels` options to set tags for the service to help identify them.
 
 !!! hint alert alert-success "Tips"
-    - When adding nodes of a sharded cluster, add each node separately using the `--cluster mycluster` option for the [MongoDB Cluster Summary](../../details/dashboards/dashboard-mongodb-cluster-summary.md) dashboard to populate correctly. Also use the `--replication-set` option to specify a replication set. Example: `--replication-set config` for your config servers; `--replication-set rs1` for your servers in the first replica set, `--replication-set rs2` for your servers in the second replica set, and so on.
+    - When adding nodes of a sharded cluster, add each node separately using the `--cluster mycluster` option for the [MongoDB Cluster Summary](../../details/dashboards/dashboard-mongodb-cluster-summary.md) dashboard to populate correctly.
```
catalinaadam, line 220:
```suggestion
    - When adding nodes to a sharded cluster, ensure to add each node separately using the `--cluster mycluster` option. This allows the [MongoDB Cluster Summary](../../details/dashboards/dashboard-mongodb-cluster-summary.md) dashboard to populate correctly. 
```

docs/setting-up/client/mongodb.md
catalinaadam, line 221:
```suggestion
    - You can also use the `--replication-set` option to specify a replication set, altough they are automatically detected. For instance, you can use `--replication-set config` for your config servers; `--replication-set rs1` for your servers in the first replica set, `--replication-set rs2` for your servers in the second replica set, and so on.
```

docs/setting-up/client/mongodb.md
```diff
@@ -217,7 +217,8 @@ Use `pmm-admin` to add the database server as a service using one of these examp
...
+    - When adding nodes of a sharded cluster, add each node separately using the `--cluster mycluster` option for the [MongoDB Cluster Summary](../../details/dashboards/dashboard-mongodb-cluster-summary.md) dashboard to populate correctly. 
+    - You can also use the `--replication-set` option to specify a replication set, altough they are automatically detected. Example: `--replication-set config` for your config servers; `--replication-set rs1` for your servers in the first replica set, `--replication-set rs2` for your servers in the second replica set, and so on.
     - Atlas doesn't support direct connections. When connecting to an Atlas instance, use the `pmm-admin` option `--direct-connection=false`. (Doing so will prevent replicaset status from working and the MongoDB Overview dashboard widget will show invalid values.)
```
catalinaadam, line 222:
```suggestion
    - When connecting to an Atlas instance, keep in mind that direct connections are not supported. Make sure to use the `--direct-connection=false` option with `pmm-admin`. Otherwise, replicaset status functionality will not work as expected, and the MongoDB Overview dashboard widget will show invalid values.
```

docs/setting-up/client/mongodb.md
BupycHuk: you are right, but it's not about can we or can't we. it's about do we or don't we. --replication-set generates label that is different from `set` label, so until we formalise and make autodetect I wouldn't mention that in documentation.

igroene: I agree. To validate if it's really needed, I ran a quick test and backups seems to work fine even without providing the --replication-set flag. Do you have any example or remember under which conditions may the backup fail?

BupycHuk: No, that was just an assumption, I checked a code and seems like we use RS for PITR, but it's read from Mongo itself.

BupycHuk (approved):

PR #1241, opened 2024-05-14
clarified more Drawbacks for PGSS

rnovikovP:


docs/setting-up/client/postgresql.md
```diff
@@ -80,7 +80,7 @@ Here are the benefits and drawbacks of each.
 
 |                      | <i class="uil uil-thumbs-up"></i> Benefits     | <i class="uil uil-thumbs-down"></i> Drawbacks
 |----------------------|------------------------------------------------|---------------------------------------------------
-| `pg_stat_statements` | 1. Part of official `postgresql-contrib` package. | 1. No aggregated statistics or histograms.<br>2. No Query Examples.
+| `pg_stat_statements` | 1. Part of official `postgresql-contrib` package. | 1. No aggregated statistics or histograms<br>2. No query examples <br>3. No query examples<br>4. No plan execution information
 | `pg_stat_monitor`    | 1. Builds on `pg_stat_monitor` features.<br>2. Bucket-based aggregation. |
```
BupycHuk, line 84: should we add to pg_stat_monitor that it allows us to have everything mentioned in pg_stat_statements drawbacks part?
TBH I like tables like this more then mentioning them in list in table
image

rnovikovP: @BupycHuk, I agree we can do a better presentation here. But first, let's solve the current question and involve Pedro for better UX while we'll work on updating Doc for Install/update

PR #1309, opened 2024-09-13
Update upgrade.md

igroene:
Improve instructions for upgrade on different platforms with examples


rnovikovP:
@catalina-adam can / should we somehow use the variable for the PMM version instead of hardcoded examples?

catalinaadam (approved):

